PR #35, opened 2020-12-23
Matching

jcarrete5:


api/models.py
```diff
@@ -1,38 +1,41 @@
...
 
-#All possible relationships
-#(true,true), (true, null), (false, true/null/false), (null,null)
-#Note that we don't need to store (null,null)
+
+# All possible relationships
+# (true,true), (true, null), (false, true/null/false), (null,null)
+# Note that we don't need to store (null,null)
 class MatchStatus(models.Model):
     primary_user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="+")
```
mneedleman: So the original version of this was actually incorrect. Since we decided to store entries when the initial query is made, there WILL be (null,null). So primary_user needs a null=True

api/models.py
```diff
@@ -1,38 +1,41 @@
...
-    secondary_user = models.ForeignKey(User, null=True, on_delete=models.CASCADE, related_name="+")
+    secondary_user = models.ForeignKey(
+        User, null=True, on_delete=models.CASCADE, related_name="+"
+    )
     primary_response = models.BooleanField()
```
mneedleman: Same here

api/models.py
```diff
@@ -1,38 +1,41 @@
...
     secondary_response = models.BooleanField(null=True)
 
     @property
     def accepted(self):
-        return self.primary_response and self.secondary_accepted()
+        return self.primary_response and self.secondary_accepted
```
mneedleman: This would also need to be updated

frontend/urls.py
```diff
@@ -1,6 +1,7 @@
-from django.urls import path
+from django.urls import re_path
+
 from . import views
 
 urlpatterns = [
-    path('', views.index)
-]
\ No newline at end of file
+    re_path("", views.index),
```
mneedleman, line 6: What did this accomplish? And do we need to consider a converter to maintain standard use of `path`, which is the preferred use

frontend/views.py
```diff
@@ -1,5 +1,6 @@
+from django.http.request import HttpRequest
 from django.shortcuts import render
 
-# Create your views here.
-def index(request):
-    return render(request, 'frontend/index.html')
\ No newline at end of file
+
+def index(request: HttpRequest):
```
mneedleman, line 5: I don't think this needs a specific type, but I'm curious to hear your thoughts

frontend/urls.py
jcarrete5: By changing this to `re_path`, all requests will be directed to the same view responding with the react app. Then react router can handle displaying the correct UI based on the URI

frontend/views.py
jcarrete5: It doesn't but it allows vscode to add better autocompletion support

frontend/urls.py
jcarrete5: It was just easier to match all URIs with a blank regex than some `path` type weirdness

frontend/views.py
jcarrete5: In general I've noticed that typing the variables results in better autocompletion. Variable declarations get typed a bit differently so sometimes you might see `... # type: <some-type>` after a declaration which lets the parser know the type of the variable

frontend/urls.py
mneedleman: Awesome, I was hoping that was the case

frontend/views.py
mneedleman: Good tip!

jcarrete5:
@Deravadom The `get_queryset` method in the `Matches` view is not returning the correct `QuerySet` and I'm struggling to figure out why. It's slightly modified from what we discussed the other day but the same idea is there.

api/migrations/0002_matchstatus.py
```diff
@@ -25,25 +26,33 @@ class Migration(migrations.Migration):
                         verbose_name="ID",
                     ),
                 ),
-                ("primary_response", models.BooleanField()),
-                ("secondary_response", models.BooleanField(null=True)),
+                ("user_lo_response", models.BooleanField(null=True)),
+                ("user_hi_response", models.BooleanField(null=True)),
```
Deravadom, line 30: Why rename it?

api/tests.py
```diff
@@ -1,3 +1,57 @@
...
+from django.contrib.auth.models import User
+from django.test import Client, TestCase
+from rest_framework.response import Response
+
+from api.models import MatchStatus
+
+
+class MatchesViewTestCase(TestCase):
+    def setUp(self):
+        ale = User.objects.create_user(username="ale", password="password")
```
mneedleman, line 9: Remind me to set up factoryboy and faker so we can do this a little easier

api/models.py
```diff
@@ -1,38 +1,66 @@
...
+        constraints = [
+            # This constraint avoids duplicate/illegal records:
+            #   user1  user2   . . .
+            #   -----  -----
+            #   0034   0035     }
+            #   0035   0034     } Only one of these two records needs to exist
+            #   . . .
+            #   0022   0022     Users can't match with themselves
+            models.CheckConstraint(
+                check=Q(user_lo_id__lt=F("user_hi_id")), name="user_lo_lt_user_hi"
```
mneedleman, line 40: Why was F needed here?

api/tests.py
```diff
@@ -1,3 +1,57 @@
...
+            user_lo=boy, user_hi=dig, user_lo_response=None, user_hi_response=True
+        )
+        MatchStatus.objects.create(
+            user_lo=boy, user_hi=eel, user_lo_response=None, user_hi_response=False
+        )
+        MatchStatus.objects.create(
+            user_lo=boy, user_hi=fog, user_lo_response=None, user_hi_response=None
+        )
+
+    def test_get_potential_matches_basic(self):
```
mneedleman: Please include some comments here

mneedleman (commented):
I still think it needs to be `ms_hi__user_hi`.

api/models.py
jcarrete5: You need to use F to reference the value of the column from the table

mneedleman: Why does `user_lo_id < user_hi_id` prevent self-matching? To me this introduces a weird side-affect

jcarrete5: Because if `user_lo_id == user_hi_id`, this would violate the constraint and not be allowed into the database

jcarrete5: It doesn't directly prevent self-matching. It's more of a database integrity thing.

mneedleman: So why isn't it `check=~Q(user_lo_id=F("user_hi_id")), name="user_lo_neq_user_hi"`

jcarrete5: Because we also don't want entries where the two user ids are swapped because they mean the same thing.

mneedleman: What do you mean by swapped?

mneedleman: Ah I understand now, you're working to avoid duplicates as well. Now this makes sense

jcarrete5:
`matchstatus_hi__user_lo` is correct

mneedleman (approved):
Great as an initial version! Fantastic work

PR #36, opened 2020-12-29
Token fixing

mneedleman:
Removed the `get`, `put`, and `post` fetch wrappers for a combined `request` wrapper.
Unfortunately, github mangled the diff for `fetch.tsx`, so it's less clear what I did.

I also updated the TokenPage to use the new request function, and made sure the jwt auth was included in the requests where needed.


mneedleman:
At some point, we might want to consider transitioning from a request function to an async hook

rao53 (approved):
looks good..

jcarrete5 (approved):

PR #41, opened 2021-01-14
Fr a1 account creation

### LINKED ISSUE #20: FR-A1: Account Creation
Note this is separate from the profile.
While there is some monstrosity of an implementation present, please update the FE to look prettier.

See FR-A5,A7,A8 for details.

rao53:
If the migrations don't work. Simply reset the database... The account creation API is much cleaner and multiple requests aren't being made now. Users and profile model are being made when you sign up with a default value for everything in the profile model which can be updated as needed.


mneedleman:
By resetting the database, do you mean deleting db.sqlite3? That works for now, but is not a good long-term solution. We should work to have migrations that don't cause issues. It may be that you just need to re-apply the migrations. If needed, I can post some documentation on how to do that properly.

Additionally, we shouldn't be overwriting or editing the generated migrations.

rao53:
> By resetting the database, do you mean deleting db.sqlite3? That works for now, but is not a good long-term solution. We should work to have migrations that don't cause issues. It may be that you just need to re-apply the migrations. If needed, I can post some documentation on how to do that properly.
> 
> Additionally, we shouldn't be overwriting or editing the generated migrations.

That is true. It was a mistake I made when I deleted earlier migrations when I was testing. Long term, it should work when migrations are re-applied.

api/models.py
```diff
@@ -1,18 +1,108 @@
...
+    age = models.PositiveIntegerField(null=True, blank=False)
+    GENDER_CHOICES = (
+        ("M", "Male"),
+        ("F", "Female"),
+    )
+    gender = models.CharField(max_length=1, choices=GENDER_CHOICES, null=True)
+
+    class Meta:
+        """
+        to set table name in database
```
jcarrete5: I'd leave this out because it is a bit redundant and not entirely accurate. The `Meta` class can do more than specify the table name.

api/serializers.py
```diff
@@ -1,39 +1,71 @@
+from django.contrib.auth import authenticate
+from django.contrib.auth.models import update_last_login
 from rest_framework import serializers
 from rest_framework_jwt.settings import api_settings
-from django.contrib.auth.models import User
-from .models import Video
+
+# from django.contrib.auth.models import User
```
jcarrete5: I really dislike commented code like this. If the `User` model is not needed, why not completely remove it?

api/views.py
```diff
@@ -1,37 +1,111 @@
...
+JWT_PAYLOAD_HANDLER = api_settings.JWT_PAYLOAD_HANDLER
+JWT_ENCODE_HANDLER = api_settings.JWT_ENCODE_HANDLER
 
-class UserList(APIView):
-    """
-    Create a new user
-    """
-    permission_classes = (permissions.AllowAny,)
+# @api_view(['GET'])
+# def current_user(request):
```
jcarrete5: Same thing. If this function is no longer needed, it should be removed.

jcarrete5 (commented):
Nice job overall. In general, I feel that removing unused code is better than commenting it out. If it turns out to be needed later, it can be restored from version control and by removing it, there are less distractions when reading.

api/models.py
```diff
@@ -1,18 +1,108 @@
...
+    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="profile")
+    first_name = models.CharField(max_length=50, unique=False)
+    last_name = models.CharField(max_length=50, unique=False)
+    phone_number = models.CharField(max_length=10, unique=True, null=True)
+    age = models.PositiveIntegerField(null=True, blank=False)
+    GENDER_CHOICES = (
+        ("M", "Male"),
+        ("F", "Female"),
+    )
+    gender = models.CharField(max_length=1, choices=GENDER_CHOICES, null=True)
```
jcarrete5, line 19: Might want to distinguish between an unspecified gender another user-specified non-listed gender. It is not clear what a null value in this column would represent.

rao53: For this, null is just a placeholder for now. it can be changed easily later

rao53:
Screen Shot 2021-01-29 at 1 08 32 PM
Screen Shot 2021-01-29 at 1 08 38 PM

newly created login and sign-up form

rao53:
Screen Shot 2021-01-29 at 2 32 20 PM
 
Cleaner landing page with barebone dashboard page attached

frontend/src/components/LoginForm.tsx
```diff
@@ -2,34 +2,49 @@ import * as React from 'react';
...
-export default function LoginForm({ handle_login }: LoginFormProps) {
+export default function LoginForm({ handle_login, errorMessage }: LoginFormProps) {
     const [username, setUsername] = React.useState<string>(null);
     const [password, setPassword] = React.useState<string>(null);
+    
+    // Clear Error message
+    React.useEffect(() => {
+        errorMessage = ""
+    });
+
```
mneedleman: This is problematic. You're passing a prop to a component and immediately overriding it.
If it's part of the component's state, use `useState`. If it's a prop, it's a constant.

frontend/src/components/SignupForm.tsx
```diff
@@ -3,29 +3,43 @@ import { UserHandlerArgs } from './LoginForm';
...
+export default function SignupForm({ handle_signup, errorMessage }: SignupFormProps) {
     const [username, setUsername] = React.useState<string>(null);
     const [password, setPassword] = React.useState<string>(null);
 
+
+    // Clear Error message
+    React.useEffect(() => {
+        errorMessage = ""
+    });
+
```
mneedleman: Same here.

mneedleman (changes requested):
Looks great! Just a couple small things and it's good to go.

frontend/src/components/LoginForm.tsx
rao53: fixed

mneedleman (approved):

mneedleman:
Since he was requested, please do not merge until @elp49 has reviewed this.

jcarrete5 (changes requested):
Run `black` and `isort` on the python code. `black . && isort .` from the `api/` directory.

jcarrete5 (approved):

elp49 (approved):
looks good

PR #42, opened 2021-01-19
Video upload and download.

### LINKED ISSUE #40: Fr-V2: Video Recording Tool
Just the base implementation with video preview. This should record the video and store when finished. This should also include a cancel button

See FR-V3,4,5,6,8,9 for details

mneedleman:
This PR packs quite a punch.
1. Added `VideoListPage` as a temporary way to view received videos.
2. Added the `RecordPage` to record videos. This will check the app context for the receiving user.
3. Added global state through context and hooks
4. Updated the video model, view, and serializer


api/serializers.py
```diff
@@ -33,7 +36,28 @@ class Meta:
...
+#There are currently a couple of issues here
+#1. The file needs a name, but not a title
 class VideoSerializer(serializers.ModelSerializer):
+
+    def create(self, validated_data):
+        data = self.context.get('data')
+        user = self.context.get('user')
+        #Data is prepended by "data:video/mp4;base64,", need to remove those 22 characters
+        #Encode converts from string to bytes
+        data_bytes = data[22:].encode('ascii')
```
jcarrete5: What if the data isn't mp4? Do we determine the prepended encoding and MIME type metadata? Have you had a look at [django rest framework parsers](https://www.django-rest-framework.org/api-guide/parsers/)?

mneedleman: `react-media-recorder` leaves the data as a Blob, which is then converted into a base64 encoded string using FileReader.

jcarrete5: I think `FileUploadParser` would be of particular interest here

jcarrete5: Ah okay I see. Why send the data over as a blob over just straight binary? Seems like converting the blob to base64 instead of a typed array needlessly increases the data being sent by ~33%

jcarrete5: Looks like there is a [file reader function](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsArrayBuffer) to read the data into an array

jcarrete5: I don't think the size difference alone warrants sending the data as raw binary over base64, but if the data is being saved as a blob, then the metadata should be saved along with it.

jcarrete5: Also, it looks like the `ContentFile` object supports binary strings directly so there is no need to encode the blob as ascii.

frontend/src/App.tsx
```diff
@@ -3,36 +3,52 @@ import { render } from 'react-dom';
...
+                    </Route>
+                    <Route path='/Profile'>
+                        <ProfilePage />
+                    </Route>
+                    <Route path='/'>
+                        <MainPage />
+                    </Route>
+                </Switch>
+            </Router>
+        </AppContext.Provider>
```
mneedleman: By wrapping the body `App.tsx` in `AppContext.Provider`, I've initialized the global state and applied it to the app context.

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -0,0 +1,81 @@
...
+        request('/api/video/', 'post', true, true, {
+            receiver: receiver,
+            data: reader.result,
+        });
+    };
+}
+
+export default function RecordPage() {
+    const [videoBlob, setVideoBlob] = React.useState<Blob>(null);
+    const [focusedUser] = useFocusedUser();
```
mneedleman: Here I access the global state context through my custom hook, `useFocusedUser`. Because it's a hook, which returns an array containing the value and a function to change the value, I can just say I want the value.

frontend/src/utils/colors.tsx
```diff
@@ -0,0 +1,8 @@
+import { colors } from '@material-ui/core';
+
+export const Colors = {
+    DEEP_RED: colors.red[900],
+    DEEP_BLUE: colors.blue[900],
+};
+
+export type ColorsType = typeof Colors['DEEP_RED'] | typeof Colors['DEEP_BLUE'];
```
mneedleman: From now on, let's define app colors here.

frontend/src/utils/context.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+// Even though this is initialized as null, using it will supply the correct type
+export const AppContext = React.createContext<AppContextType>({
+    focusedUser: null,
+});
+
+export default function useFocusedUser() {
+    const context = React.useContext(AppContext);
+    return context.focusedUser;
+}
```
mneedleman: Ok there's a lot going on here.
At the base level, I create a context called `AppContext`. By using `<AppContext.Provider value={something}>` I initialize the context, which is serving as our global state.

Then I define a custom hook, `useFocusedUser` to let us use the context without having to pass it down the entire component hierarchy as props.

Lastly, at the top I explicitly define the type of the context. Without this, it will be of type "any". Now it's not.

frontend/src/utils/colors.tsx
jcarrete5: I think we would be better off utilizing material ui themes for this stuff. They can automatically apply colors to mui components without needing to explicitly style them.

mneedleman: I don't have a strong opinion one way or the other

api/serializers.py
mneedleman: Let's revisit this discussion this week

frontend/src/utils/colors.tsx
jcarrete5: This also doesn't really provide any abstraction of colors. For example, if I were to style a bunch of components using `DEEP_RED` and later decide I want them to be a pink, I now need to change the color on all the components to a new color. Alternatively, I could update the `DEEP_RED` color to pink but then the variable name would no longer reflect the actual color and may break other components using `DEEP_RED`. MUI theme avoids this problem by choosing names for the theme variables that reflect their use: https://material-ui.com/customization/theming/#api . This also has the consequence of making it really easy to change the color scheme of the entire application which I believe was a feature mentioned very early on during planning. I don't remember if we really cared about this feature that much though.

mneedleman: I'll create a task to standardize coloring, since it's somewhat lower priority than the rest of the PR. But you make some good points, and I agree with you

api/models.py
```diff
@@ -4,10 +4,18 @@
...
+        User, on_delete=models.CASCADE, related_name="sent_videos"
+    )
+    receiver = models.ForeignKey(
+        User, on_delete=models.CASCADE, related_name="received_videos"
+    )
+
+    created_at = models.DateTimeField()
     viewed_at = models.DateTimeField(null=True)
 
+    video_file = models.FileField(upload_to="media/", null=True, verbose_name="")
```
jcarrete5: `upload_to="media/"` is redundant. This would put a subdirectory `media/` under a directory that is already called `media/`. Maybe a better name would be `videos/`

And do we really want to allow a null video? What purpose would a null video serve?

api/serializers.py
```diff
@@ -35,8 +39,46 @@ class Meta:
...
+        # receiver = User.objects.get(id=validated_data.pop('receiver'))
+        # Data is prepended by "data:video/mp4;base64,", need to remove those 22 characters
+        # Encode converts from string to bytes
+        data_bytes = data[22:].encode("ascii")
+        instance = self.Meta.model(**validated_data)
+
+        instance.sender = sender
+
+        # We need to do this before save because we use it for fname
+        instance.created_at = now()
```
jcarrete5, line 52: Have you considered making the default value in the model `datetime.now`?

api/serializers.py
```diff
@@ -35,8 +39,46 @@ class Meta:
...
+
+        # ok we need to construct a filename
+        fname = "%d_%d_%s" % (sender.id, instance.receiver.id, instance.created_at)
```
jcarrete5, line 55: I'd recommend setting the filename to the hash of the file. Let the metadata about the file live in the database. iirc file name collisions shouldn't be an issue because the file field will automatically append a slug to the file name if it conflicts.

api/serializers.py
```diff
@@ -35,8 +39,46 @@ class Meta:
...
-
 class VideoSerializer(serializers.ModelSerializer):
+    def create(self, validated_data):
+        data = self.context.get("data")
+        sender = self.context.get("user")
+
+        # receiver = User.objects.get(id=validated_data.pop('receiver'))
+        # Data is prepended by "data:video/mp4;base64,", need to remove those 22 characters
+        # Encode converts from string to bytes
+        data_bytes = data[22:].encode("ascii")
```
jcarrete5: We discussed this but for the purpose of just having files uploading this is fine. But I don't see why we are stripping the scheme part of the data URL. You may as well just leave it there and avoid having to tack it on later.

api/serializers.py
```diff
@@ -35,8 +39,46 @@ class Meta:
...
+
+    def to_representation(self, instance):
+        rep = super().to_representation(instance)
+
+        video_file = instance.video_file
+        video_file.open()
+
+        # Raw base64 bytes need to be decoded to a string
+        # and the data needs the header
+        rep["video_file"] = "data:video/mp4;base64," + video_file.read().decode("UTF-8")
```
jcarrete5: The file is encoded as ascii above but decoded here as UTF-8. They should be compatible because base64 encoding doesn't use any characters outside of ascii but why have it UTF-8 here and ascii above? Also worth noting that `str.encode` and `bytes.decode` both use utf-8 by default so no need to specify.
```suggestion
        rep["video_file"] = video_file.read().decode()
```

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    if (!stream) {
+        return null;
+    }
+    return <video ref={ref} className={classes.video_root} autoPlay />;
+}
+
+function sendVideo(blob: Blob, receiver: number) {
+    const reader = new FileReader();
+    reader.readAsDataURL(blob);
```
jcarrete5: You can avoid using a `FileReader` altogether. The `Blob` interface allows you to get a `ReadableStream` ([`Blob.stream`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream)) which can be used in the body of a [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request) to send the file over as binary instead of as a data URL.

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -0,0 +1,81 @@
...
+    reader.onloadend = () => {
```
jcarrete5: Prefer [`reader.load`](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/load_event) over [`reader.loadend`](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/loadend_event) here because loadend will be called even if the file read fails whereas load only gets called after a successful file read.

frontend/src/FullPageRoutes/VideoListPage.tsx
```diff
@@ -0,0 +1,33 @@
...
+    });
+
+    return (
+        <ul>
+            {videos.map(({ id, title, video_file }) => {
+                return (
+                    <li key={'video' + id}>
+                        {title}
+                        <video width={500} height={500} controls>
+                            <source type='video/mp4' src={video_file} />
```
jcarrete5: I see the appeal of having a data URL in this scenario. Ultimately, the file serving probably shouldn't even be handled by the REST API backend and instead by a proper web server.

In the meantime, instead of sending the video file back, you could send the location of the file and use that as the source of the video. Django should allow you to specify the URL to the media in debug mode. Something like `localhost:8000/media/<video_file>`. Of course when you send the file location back, you are really specifying the whole URL so this code here would be largely untouched.

frontend/src/utils/colors.tsx
jcarrete5: I'd argue that getting the theme stuff setup early should be fairly high priority. It will be annoying making the change later when a lot of the components have local styles that may or may not be cohesive with the rest of the site.

Perhaps we get a separate pull request going for setting up material UI theme context.

Edit: I find myself disagreeing with myself here haha. Components really shouldn't be concerned with css at all really unless it is layout related at this point. I do still think that this file is unnecessary for the same reasons mentioned in the previous [comments](https://github.com/Conteo-Inc/conteo/pull/42#discussion_r568021790)

api/models.py
mneedleman: Ah I hadn't noticed the duplicate directory! Nice catch.
And nope! That should not be there 👍

api/serializers.py
mneedleman: Do you mean with `auto_now_add=True`? That was not working for me because that is applied when `save()` is called.

api/serializers.py
mneedleman: I'm not sure I understand what the benefit is by doing it that way. As far as I know, the only metadata the video comes with is the file type and the size(?). But assuming there is a benefit, I'd be happy chalking it up to technical debt.

api/serializers.py
jcarrete5: Yes that is what I mean. Couldn't recall the name of the argument to set. As long as the date is used in the name of the file, I suppose it makes sense to set this manually.

mneedleman: Yeah, so that might change if we go a different naming strategy. TBH the name doesn't matter a ton, although I can see the value of metadata at some point

api/serializers.py
mneedleman: Yup this is silly

api/serializers.py
jcarrete5: By metadata, I was referring to information about the file already stored in the database. I'm not sure of any technical benefit. My thinking was by setting the filename to the  file hash, the files would have more unique names and can be spread out on a file system. Might be useful to verify the integrity of the file as well. The file names would also be hard to guess which would add more protection against unwanted file access by unauthorized users.

conteo/settings.py
```diff
@@ -1,150 +1,175 @@
...
+
+For more information on this file, see
+https://docs.djangoproject.com/en/3.1/topics/settings/
+
+For the full list of settings and their values, see
+https://docs.djangoproject.com/en/3.1/ref/settings/
+"""
+from datetime import timedelta
+from pathlib import Path
+
```
jcarrete5, line 14: Missing import breaks the settings module
```suggestion
import os

```

api/serializers.py
```diff
@@ -1,43 +1,61 @@
...
+JWT_PAYLOAD_HANDLER = api_settings.JWT_PAYLOAD_HANDLER
+JWT_ENCODE_HANDLER = api_settings.JWT_ENCODE_HANDLER
 
 
 class UserSerializer(serializers.ModelSerializer):
     class Meta:
-        model = User
-        fields = ("username",)
+        model = UserProfile
+        fields = ("first_name", "last_name", "phone_number", "age", "gender")
```
jcarrete5: This is a breaking change and causes the tests to fail. The matching view is expecting this to serialize a `User`, not a `UserProfile`. Either change the matching view to use `UserProfile` and fix the test or leave the `UserSerializer` as-is and make a `UserProfileSerializer`. I was planning on making the switch to `UserProfile`s once they were added since they have the actual data I need to construct the matches on the frontend.

jcarrete5 (changes requested):

conteo/settings.py
mneedleman: 👍

mneedleman:
This is currently blocked by #52

api/serializers.py
jcarrete5: Unclear if you are going to leave in the removal of the 22 chars. I don't understand why they need to be removed only to be added in again later.
```suggestion
        data_bytes = data.encode()
```

jcarrete5 (changes requested):

mneedleman:
Odd that I cant reply, but I'm looking into changing that

jcarrete5:
> Odd that I cant reply, but I'm looking into changing that

You can reply above where the original review was started. Kind of confusing

rao53 (changes requested):
Could you possibly update this branch with the master?

rao53:
Just to confirm, the major objective to this PR is to be able to record and view a video?

rao53:
I also noticed that after I record a video, my webcam is still on regardless of the fact that I am not on the record page...

mneedleman:
Yes the major objective is to record, send, and view videos. Also I still need to fix the video view to only get received videos. And yes the webcam issue is a known bug. It'll take a decent amount of time to address it, unfortunately

rao53:
> Yes the major objective is to record, send, and view videos. Also I still need to fix the video view to only get received videos. And yes the webcam issue is a known bug. It'll take a decent amount of time to address it, unfortunately

Gotcha.. Understandable

api/serializers.py
```diff
@@ -1,74 +1,49 @@
...
         data = self.context.get("data")
         sender = self.context.get("user")
 
-        # receiver = User.objects.get(id=validated_data.pop('receiver'))
-        # Data is prepended by "data:video/mp4;base64,",
-        # need to remove those 22 characters
+        # Data is prepended by "data:video/mp4;base64,"
         # Encode converts from string to bytes
-        data_bytes = data[22:].encode("ascii")
+        data_bytes = data.encode("ascii")
```
jcarrete5: Still unclear as to why this is specifically being encoded with ascii. There is no benefit to encoding with ascii over utf-8. And you are decoding as utf-8 later.
```suggestion
        data_bytes = data.encode()

```

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -1,81 +1,79 @@
...
+    return null
+  }
+  return <video ref={ref} className={classes.video_root} autoPlay />
+}
+
+function sendVideo(blob: Nullable<Blob>, receiver: NullableId) {
+  if (blob) {
+    const reader = new FileReader()
+    reader.readAsDataURL(blob)
+    reader.onloadend = () => {
```
jcarrete5: [Consider using `onload` here](https://developer.mozilla.org/en-US/docs/Web/API/FileReader#events). `onloadend` will be called even if there is an error.

frontend/src/FullPageRoutes/VideoListPage.tsx
```diff
@@ -1,33 +1,34 @@
...
+  id: number
+  title: string
+  video_file: string
+}
+
+export default function VideoListPage(): JSX.Element {
+  const [videos, setVideos] = React.useState<Video[]>([])
+
+  React.useEffect(() => {
+    request<Video[]>("/api/video/", "get", true, false).then((res) => {
```
jcarrete5: Check your arguments here; I don't think you are intending to specify false as the body of the request.

frontend/src/utils/fetch.tsx
```diff
@@ -9,7 +9,7 @@ export async function http<T>(request: RequestInfo): Promise<HttpResponse<T>> {
 
   try {
     res.parsedBody = await res.json()
-  } catch (ex) {}
+  } catch (ex) {} //eslint-disable-line
```
jcarrete5: I don't think ignoring this error is a good solution. Removing this try-catch and always parsing the body as json would solve all of the issues involving null checks on the parsed body. At the very least, log a message with the exception so we know when this happens whether the caller was expecting json or not. Or even better, don't assume the response is json and let the caller decide how they want to parse the body.

jcarrete5 (changes requested):

jcarrete5 (changes requested):
Only thing missing is a test for the created video list view.

jcarrete5 (dismissed):

api/tests.py
```diff
@@ -151,3 +151,41 @@ def test_get_matches_basic(self):
...
+
+        # match them together
+        MatchStatus.objects.create(
+            user_lo=self.sender,
+            user_hi=self.receiver,
+            user_lo_response=True,
+            user_hi_response=True,
+        )
+        self.video = "video"
+
```
AyalaCnaan, line 170: Not sure if this matters, but what if user_hi is sending to user_lo?

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -0,0 +1,79 @@
...
+    status,
+    startRecording,
+    stopRecording,
+    mediaBlobUrl,
+    previewStream,
+  } = useReactMediaRecorder({
+    video: true,
+    onStop: (_, blob) => setVideoBlob(blob),
+  })
+
```
AyalaCnaan, line 60: This is tres elegant.

AyalaCnaan (dismissed):
Looks like a great work in progress to me. Good to move on to further tasks as far as I'm concerned.

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -0,0 +1,79 @@
+import * as React from "react"
```
rao53, line 1: We probably want to make sure a user is logged in before they can record a video but I'm guessing for this PR, it's not required.

rao53 (approved):

mneedleman:
Since both @AyalaCnaan and @jcarrete5 wanted small changes that have been addressed, I'm calling this good

PR #43, opened 2021-01-22
Dashboard 

sb3958:


frontend/src/FullPageRoutes/Dashboard.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { BrowserRouter as Router, Link, Route, Switch } from 'react-router-dom';
+import HomeIcon from '@material-ui/icons/Home';
+import Person from '@material-ui/icons/Person';
+import ListAlt from '@material-ui/icons/ListAlt';
+import Mail from '@material-ui/icons/Mail';
+import Info from '@material-ui/icons/Info';
+import Help from '@material-ui/icons/Help';
+
+//Not sure how to put these in, we should look into it:
+// @media (prefers-reduced-motion: no-preference) {
```
jcarrete5: This looks like the starter react code. Just delete this.

jcarrete5 (changes requested):
Looks nice!


image

frontend/App.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
```
mneedleman: This file is in the wrong location.

frontend/App.tsx
```diff
@@ -0,0 +1,41 @@
...
+import LinkItem from './components/LinkItem';
+import TokenPage from './FullPageRoutes/TokenPage';
+import DashboardPage from './FullPageRoutes/DashboardPage'
+
+
+function MainPage() {
+    return (
+        <ul>
+            <LinkItem to='/Tokens' text='Tokens' />
+            <LinkItem to='/Dashboard' text='Dashboard' />
```
mneedleman: It looks like you're behind, this is missing the profile. You need to pull

frontend/App.tsx
```diff
@@ -0,0 +1,41 @@
...
+                <Route exact path='/Dashboard'>
+                    
+                    <DashboardPage />
+                </Route>
+                <Route path='/Tokens'>
+                    <TokenPage />
+                </Route>
+                <Route path='/'>
+                    <MainPage />
+                </Route>
```
mneedleman: I think it's safe to say that all of these should be exact. I'm not 100% certain about the `/` path. But let's try it

frontend/DashboardPage.tsx
```diff
@@ -0,0 +1,177 @@
+import * as React from 'react';
```
mneedleman: This file is in the wrong location.

frontend/DashboardPage.tsx
```diff
@@ -0,0 +1,177 @@
...
+   
+
+});
+
+
+export default function DashboardPage() {
+    const classes = useStyles();
+
+    // Add the options for the sidebar to the Dashboard list.
+    var options: DashboardItems[] = [
```
mneedleman: This should be a `const`

frontend/DashboardPage.tsx
```diff
@@ -0,0 +1,177 @@
...
+            icon: <EditIcon />,
+            optionsName: 'Edit Profile',
+        },
+        {
+            icon: <InfoIcon />,
+            optionsName: 'Tutorials',
+        },
+    ];
+
+    var contents: Maillist[] = [
```
mneedleman: This should also be a const

frontend/DashboardPage.tsx
```diff
@@ -0,0 +1,177 @@
...
+    optionsName: string;
+}
+
+type Maillist = {
+    profile: JSX.Element;
+    name: string;
+    message: string;
+    dateReceived: string;
+
+}
```
mneedleman: I think this type name is misleading. The values describe a single item, but `Maillist` describes a list. The L should be capitalized.

frontend/DashboardPage.tsx
```diff
@@ -0,0 +1,177 @@
...
+            dateReceived: 'Oct 18',
+        },
+        
+    ];
+
+
+    return (
+        
+    <>
+        <div style={{ margin: 0, padding: 0, boxSizing: 'border-box'}}>
```
mneedleman: Let's avoid inline styling as much as possible. Sometimes it is unavoidable, such as with external libraries. Also, you should be using a MUI Box here.

frontend/DashboardPage.tsx
```diff
@@ -0,0 +1,177 @@
...
+                    <LinkItem to='/ContactUs' text='Contact Us' />
+                    <LinkItem to='/Help' text='Help' />
+                    <LinkItem to='/LogOut' text='Log Out' />
+                    
+                </ul>
+            </nav>
+        </div>
+
+        <div className={classes.mailSection}>
+        <Grid container>
```
mneedleman: It looks like your spacing here is a bit off. Also it doesn't make sense to wrap a grid in a div. Grid has a `className` prop, so you can apply the css to it.

frontend/DashboardPage.tsx
```diff
@@ -0,0 +1,177 @@
...
+        {options.map(({ icon, optionsName }: DashboardItems) => (
+            <Grid key={optionsName} container item className={classes.bar} xs={6}>
```
mneedleman: I'm not sure what happens if you specify that a Grid is both a container and an item. Can you explain?

frontend/DashboardPage.tsx
```diff
@@ -0,0 +1,177 @@
...
+                <Grid container item alignItems="center" justify="center" xs={3}>
```
mneedleman: However, I am fairly certain that a Grid should only be a container if it has other Grids as children.

mneedleman (changes requested):
First and foremost you need to make sure that your files are in the right place. I'm not sure how they were misplaced, but I suspect is has something to do with how you submitted the work. We should always be using git, not direct upload/download.

mneedleman:
image

api/models.py
```diff
@@ -2,18 +2,16 @@
 from django.db import models
 from django.db.models import F, Q
 
+GENDER_CHOICES = (("M", "Male"), ("F", "Female"), ("O", "Other"))
```
jcarrete5: I'd say it is better to leave the choices as a class variable in the Profile because they are associated with the profile. [Django also recommends this](https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.Field.choices). 
> Though you can define a choices list outside of a model class and then refer to it, defining the choices and names for each choice inside the model class keeps all of that information with the class that uses it, and helps reference the choices (e.g, Student.SOPHOMORE will work anywhere that the Student model has been imported)

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -30,24 +30,16 @@ type UserProfile = {
...
-        )
-      }
+      setUsername(profile.parsedBody ? profile.parsedBody.first_name : "")
     })
   })
 
   // Dummy data.
   const content = {
-    username: username,
+    username: username ? username : "",
```
jcarrete5: Consider using [null coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing) here. Not that important considering this is just dummy data but I thought I'd mention it here since typescript doesn't do that great a job making it obvious that this exists.
```suggestion
    username: username ?? "",
```

frontend/src/FullPageRoutes/TokenPage.tsx
```diff
@@ -31,7 +31,7 @@ const useStyles = makeStyles({
 export default function TokenPage(): JSX.Element {
   const [displayedForm, setDisplayedForm] = React.useState<string | null>(null)
   const [logged_in, setLoggedIn] = React.useState<boolean>(false)
-  const [email, setEmail] = React.useState<string | null>(null)
+  const [, setEmail] = React.useState<string | null>(null)
```
jcarrete5: Is this a typo? Where is the first value?
```suggestion
  const [email, setEmail] = React.useState<string | null>(null)
```

frontend/src/components/dashboard/MailList.tsx
```diff
@@ -0,0 +1,26 @@
...
+import * as React from "react"
+import { request } from "../../utils/fetch"
+import MailItem, { MailListItem } from "./MailItem"
+
+export default function MailList(): JSX.Element {
+  const [mail, setMail] = React.useState<MailListItem[]>([])
+
+  React.useEffect(() => {
+    request<MailListItem[]>("/api/profiles/", "get", true).then((mailList) => {
+      setMail(mailList.parsedBody ? mailList.parsedBody : [])
```
jcarrete5: Same null coalescing suggestion stated in an earlier comment. I think your recent changes to parsed body might make this a non-issue. 
```suggestion
      setMail(mailList.parsedBody ?? [])
```

frontend/src/FullPageRoutes/TokenPage.tsx
```diff
@@ -42,47 +42,31 @@ export default function TokenPage(): JSX.Element {
...
-      setDisplayedForm("login")
-    })
-  }
+  // const handle_logout = () => {
+  //   request("/api/logout/", "post", true).then(() => {
+  //     setLoggedIn(false)
+  //     setEmail(null)
+  //     setDisplayedForm("login")
+  //   })
+  // }
```
jcarrete5: You know how I feel about commented code blocks. Why are we not doing logout anymore? Are we just going to re-implement it somewhere else?
```suggestion
```
```

File: frontend/src/FullPageRoutes/TokenPage.tsx [hunk 18]
```diff
@@ -91,7 +75,9 @@ export default function TokenPage(): JSX.Element {
   return (
     <>
       {logged_in === true ? (
-        <Dashboard handle_logout={handle_logout} email={email} />
+        // <Dashboard handle_logout={handle_logout} email={email} />
```
jcarrete5:
```suggestion
```

jcarrete5 (changes requested):
With factoryboy and faker, we need to either have a set seed to use when generating data or print the seed during test runs for reproducibility. I don't know how or where to do this within factoryboy/faker but we need something there

frontend/DashboardPage.tsx
elp49: I'm not completely certain if this is right, but from testing it I think it just saves you from adding an additional Grid. As long as all of it's immediate children are items then it could potentially save on cluttering jsx

frontend/DashboardPage.tsx
elp49: I don't think having a Grid with container and item does anything here when it's parent is also container item

api/views.py
```diff
@@ -30,14 +32,17 @@ def post(self, request: request.Request):
...
+        request: request.Request,
+    ):
         username = request.data["username"]
         password = request.data["password"]
 
         user = authenticate(request=request, username=username, password=password)
         if user is not None:
             login(request=request, user=user)
-            return response.Response(status=status.HTTP_200_OK)
+            return HttpResponseRedirect(redirect_to="/dashboard/")
```
elp49: I was fiddling around with the login bug and found that this is where the problem is.
The redirect is a good idea, but I tested it and I can't get it to redirect. It only stays on the Token page after login. have you had different results?

api/models.py
mneedleman: 👍

frontend/src/FullPageRoutes/Profile.tsx
mneedleman: 🍾 🍷 👍

frontend/src/FullPageRoutes/TokenPage.tsx
mneedleman: This is not a typo. `email` is not used, so it doesn't need to be assigned.

frontend/src/components/dashboard/MailList.tsx
mneedleman: 👍

api/views.py
mneedleman: Yeah I've been able to redirect. I'll look into it more

frontend/src/FullPageRoutes/TokenPage.tsx
jcarrete5: Okay so if email is never used, then why is there a state for it? Will it be used later but right now, it is not being used?

AyalaCnaan: There are definitely things email will be used for. But we probably haven't gotten to them yet.

mneedleman: > 
> 
> Okay so if email is never used, then why is there a state for it? Will it be used later but right now, it is not being used?

That's a good point. However, more fun to come that will make this obsolete

mneedleman:
> 
> 
> With factoryboy and faker, we need to either have a set seed to use when generating data or print the seed during test runs for reproducibility. I don't know how or where to do this within factoryboy/faker but we need something there

The more I think about this, the less sure I am that there should be a set seed. Let's discuss on Monday during our meeting

frontend/src/FullPageRoutes/TokenPage.tsx
mneedleman: Other PRs will fix this

jcarrete5:
> > With factoryboy and faker, we need to either have a set seed to use when generating data or print the seed during test runs for reproducibility. I don't know how or where to do this within factoryboy/faker but we need something there
> 
> The more I think about this, the less sure I am that there should be a set seed. Let's discuss on Monday during our meeting

There _needs_ to be some way to reproduce an error in the case where the test fails. There should either be a set seed or the seed used should be printed out during the test.

Debugging an error without having the exact input that caused the error is difficult and frustrating especially when we can easily make the error reproducible by knowing the seed.

It looks like [faker defaults to seeding the PRNG with a value of 0](https://faker.readthedocs.io/en/master/pytest-fixtures.html?highlight=seed#seeding-configuration). To be safe, we should confirm this behaviour.

elp49 (dismissed):

api/factory.py
```diff
@@ -0,0 +1,68 @@
...
+        model = User
+
+    class Params:
+        pw = "ls"
+
+
+class MatchStatusFactory(factory.django.DjangoModelFactory):
+    @lazy_attribute
+    def user_lo_response(self):
+        return random.choice(self.response_choices)
```
jcarrete5, line 57:
```suggestion
        return factory.random.choice(self.response_choices)
```

api/factory.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @lazy_attribute
+    def user_hi_response(self):
+        return random.choice(self.response_choices)
```
jcarrete5, line 61:
```suggestion
        return factory.random.choice(self.response_choices)
```

api/factory.py
```diff
@@ -0,0 +1,68 @@
...
+    def last_name(self):
+        return self.full_name.split(" ")[1]
+
+    phone_number = factory.Faker("phone_number")
+    user = factory.SubFactory("api.factories.UserFactory", profile=None)
+    birth_date = factory.Faker("date")
+
+    @factory.lazy_attribute
+    def gender(self):
+        return random.choice(self.gender_choices)[0]
```
jcarrete5, line 26:
```suggestion
        return factory.random.choice(self.gender_choices)[0]
```

api/views.py
```diff
@@ -93,6 +98,11 @@ def post(self, request, format=None):
             return response.Response(serializer.data, status=status.HTTP_201_CREATED)
         return response.Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+    def get(self, request):
+        queryset = self.filter_queryset(self.get_queryset())
+        serializer = self.list_serializer_class(queryset, many=True)
```
jcarrete5: Why is list_serializer_class used instead of just using the VideoListSerializer directly?

frontend/src/App.tsx
```diff
@@ -59,5 +76,7 @@ export default function App(): JSX.Element {
   )
 }
 
+// export default App;
+
```
jcarrete5:
```suggestion
```

jcarrete5 (changes requested):

PR #44, opened 2021-01-24
Starting to make the base UI

mneedleman:
image

We can work on spacing later.


rao53 (commented):
I believe you should separate the Header and Footer into separate components. It would make it easier to use across various pages

frontend/src/App.tsx
```diff
@@ -13,16 +36,69 @@ function MainPage() {
...
+                className={classes.header}
+            >
+                <Typography variant='h5' className={classes.bannerText}>
+                    {'Hi, Jane'}
+                </Typography>
+                {/* Figure out sizes */}
+                <Grid item lg={2} sm={2} xs={2}>
+                    <Grid container direction='row' justify='space-between'>
+                        <Typography className={classes.bannerText}>
+                            {'About'}
```
jcarrete5, line 58: Why is text embedded as javascript string literals?

frontend/src/App.tsx
```diff
@@ -13,16 +36,69 @@ function MainPage() {
...
+                direction='row'
+                justify='space-between'
+                alignItems='center'
+                className={classes.header}
+            >
+                <Typography variant='h5' className={classes.bannerText}>
+                    {'Hi, Jane'}
+                </Typography>
+                {/* Figure out sizes */}
+                <Grid item lg={2} sm={2} xs={2}>
```
jcarrete5, line 55: I think you can get away with only specifying the smallest breakpoint (xs) here and the larger values will default to that value unless specified but I'm having trouble finding the reference.

jcarrete5:
> I believe you should separate the Header and Footer into separate components. It would make it easier to use across various pages

The header and footer already do this because each page will be a route specified in the switch element and thus will have the header and footer above and below.

frontend/src/App.tsx
mneedleman: That's a good point. I'll double check that

frontend/src/App.tsx
mneedleman: I just prefer it that way

mneedleman: I just prefer it that way

jcarrete5 (approved):

jcarrete5:
This can be merged, what are we waiting for? The code has been reviewed by every reviewer and approved by at least 1.

jcarrete5 (approved):

PR #46, opened 2021-02-01
Matching UI

jcarrete5:


api/models.py
```diff
@@ -3,11 +3,7 @@
 from django.db.models import F, Q
 
 
-# Note that enforcement on these fields should be done on the frontend --- Very easy!
-class Profile(models.Model):
-    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="profile")
-    first_name = models.CharField(max_length=50, blank=True)
-    last_name = models.CharField(max_length=50, blank=True)
+class Profile(User):
```
mneedleman: This is a good idea

api/serializers.py
```diff
@@ -1,36 +1,27 @@
...
+        fields = (
+            "id",
+            "username",
+            "email",
+            "first_name",
+            "last_name",
+            "phone_number",
+            "age",
+            "gender",
+        )
```
mneedleman: Nice

api/tests.py
```diff
@@ -1,18 +1,18 @@
...
-        eel = User.objects.create_user(username="eel", password="password")
-        fog = User.objects.create_user(username="fog", password="password")
-        User.objects.create_user(username="gil", password="password")
+        ale = Profile.objects.create_user(username="ale", password="password")
+        boy = Profile.objects.create_user(username="boy", password="password")
+        cad = Profile.objects.create_user(username="cad", password="password")
+        dig = Profile.objects.create_user(username="dig", password="password")
+        eel = Profile.objects.create_user(username="eel", password="password")
+        fog = Profile.objects.create_user(username="fog", password="password")
+        Profile.objects.create_user(username="gil", password="password")
```
mneedleman: Let's chat about testing after this is merged

api/tests.py
```diff
@@ -50,77 +50,51 @@ def test_get_matches_basic(self):
         user, users who have not declined the requesting user, and users
         for which the requesting user has not already made a decision for.
         """
+
+        def usernames(data: list[dict]):
+            for profile in data:
+                yield profile["username"]
```
mneedleman: Seems like a good idea to call in the cavalry: 🐼

api/views.py
```diff
@@ -101,7 +96,8 @@ def get_queryset(self):
 
     def get(self, request: request.Request):
         max_amount = request.query_params.get("amount", 20)
-        queryset = self.filter_queryset(self.get_queryset())
-        serializer = self.get_serializer(queryset, many=True)
-        matches = random.sample(serializer.data, min(max_amount, len(serializer.data)))
-        return response.Response(matches)
+        response = self.list(request)
+        response.data = random.sample(
+            response.data, min(max_amount, len(response.data))
```
mneedleman: 🐼

mneedleman (changes requested):
I know this is still a draft, but here are some things to consider. Also, this seems like a huge vertical slice, so be careful. Let's chat about how this affects our timeline

jcarrete5:
> I know this is still a draft, but here are some things to consider. Also, this seems like a huge vertical slice, so be careful. Let's chat about how this affects our timeline

Not sure what you mean by "huge vertical slice".

jcarrete5:
I haven't implemented updating match statuses on the backend. That might be better suited for another PR.

api/serializers.py
AyalaCnaan: Why remove the "__all__" in favor of individual fields? Are there specific fields not included, and if so, what's the reason?

AyalaCnaan (dismissed):
I'm going to admit straight up that a lot of this out of my realm of comprehension. That said, I have a three comments:
1. Generally if I see a pull request for something with UI I expect to see a screenshot or several. Just to get a sense of what's happening. Not a big deal this time, just in general. 
2. I'm good with the 'onAccept' and other functions in Matching.tsx being a TODO, but we should probably state a date for the TODO. 
3. Also in the TODO category is some sort of trigger to load the next 20 profiles, which I don't think exists yet (though I admit again I don't understand everything that I'm looking at here). 

Overall I'm cool with this and impressed at the cleanliness of much of the work. It's elegant.

api/serializers.py
mneedleman: Oh I see, this excludes `is_superuser`, `is_staff`, and `is_active`. I suppose you could exclude them specifically, but it's not a big deal either way.

frontend/src/FullPageRoutes/Matching.tsx
```diff
@@ -0,0 +1,76 @@
...
+          </Button>
+          <Button onClick={() => onReport(match.id)}>
+            <Flag />
+            Report
+          </Button>
+        </>
+      ) : (
+        <></>
+      )}
+    </Container>
```
mneedleman: If you pull the conditional outside of container, you can avoid using the fragment altogether/

mneedleman (changes requested):
Just a couple small things, but overall looks great!

api/serializers.py
jcarrete5: I was also going off of [DRF's recommendation](https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include). I'm favoring being explicit about what we want to get from the serializer.

api/models.py
rao53: This implementation is actually much cleaner! Good job...

api/tests.py
```diff
@@ -50,77 +50,51 @@ def test_get_matches_basic(self):
...
-            [
-                {"username": "ale"},
-                {"username": "boy"},
-                {"username": "cad"},
-                {"username": "dig"},
-                {"username": "eel"},
-                {"username": "fog"},
-            ],
+            usernames(res.json()),
+            ("ale", "boy", "cad", "dig", "eel", "fog"),
```
rao53: Not sure if this is important for this PR but I tried running the test and the below is what I got:
Screen Shot 2021-02-18 at 7 30 11 PM

frontend/src/FullPageRoutes/Matching.tsx
jcarrete5: Well the component needs to return something to render and I figure there will probably be changes made to this body later anyway for stylistic reasons. If I pull the condition out, I still will need the else clause with the fragment.

api/tests.py
jcarrete5: This is a problem with python 3.8. [Python 3.9](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections) allows type hints to use the respective collection now. Previously, you had to import a special object from the typings module.

api/views.py
```diff
@@ -83,12 +92,17 @@ def post(self, request, format=None):
...
     View for requesting new matches for a user.
     """
 
-    serializer_class = UserSerializer
+    def get_serializer_class(self):
+        if self.action == "list":
+            return ProfileSerializer
+        if self.action == "partial_update":
+            return MatchStatusSerializer
+        return super().get_serializer_class()
```
mneedleman: To me, this is a signal to use two separate endpoints and views, but it's not a big deal either way.

mneedleman (approved):
This looks good to me

api/views.py
jcarrete5: I don't really like it much either. If I move the matches from the matchstatus model to many-to-many on the profile, then I could probably avoid this since I'd be able to use the same profile serializer for updating and getting matches.

I spent a good amount of time thinking about how to do this best. I don't think there should be two endpoints because I view the /api/matches/ endpoint as referring to a collection of matches that can be operated on. So naturally, getting matches and updating matches should both be operating on the matches collection, regardless of how the collection is managed on the backend.

api/migrations/0005_update_profile.py
```diff
@@ -0,0 +1,57 @@
...
+            name="id",
+        ),
+        migrations.RemoveField(
+            model_name="profile",
+            name="last_name",
+        ),
+        migrations.RemoveField(
+            model_name="profile",
+            name="user",
+        ),
```
AyalaCnaan, line 42: Explain this to me like I'm in first grade: Why is this being removed and what happens to these fields?

api/tests.py
```diff
@@ -77,80 +77,88 @@ def test_get_matches_basic(self):
...
+        )
+        self.assertEqual(res.status_code, 200)
+        obj = MatchStatus.objects.get(user_lo_id=1, user_hi_id=2)
+        self.assertEqual(obj.user_hi_response, False)
+        res = self.client.put(
+            "/api/matches/", {"matchId": 1, "response": True}, format="json"
+        )
+        self.assertEqual(res.status_code, 200)
+        obj = MatchStatus.objects.get(user_lo_id=1, user_hi_id=2)
+        self.assertEqual(obj.user_hi_response, True)
```
AyalaCnaan: I'm just clarifying for myself (because I'm learning a lot of this) what is happening in this test. The test is to see how the system responds when one or both users respond to the match?

AyalaCnaan (commented):
I don't know enough to say that this is ready to go or not. Seems reasonable to me.

api/migrations/0005_update_profile.py
jcarrete5: Because this is a generated file, it's usually easier to talk about the changes on the actual model in models.py which is what ultimately creates this file.

Anyway, these fields in the profile are redundant because they already exist within the user model. By extending the user model, we get the functionality we want for free. The id and user fields are unneeded because the implicit user_ptr field can double as a primary key and a reference to the related user. The first and last name exist in the user model and don't need to be repeated in the profile.

AyalaCnaan: Thanks!

api/tests.py
jcarrete5: Yea its just testing that the endpoint will update the match status table to reflect the new state of the match between two users.

AyalaCnaan: Okay, does it check what happens if only one person responds? Sorry, not great yet at reading the code.

AyalaCnaan (dismissed):

api/tests.py
jcarrete5: It checks that a user_lo and user_hi can both accept and reject a match.

mneedleman (changes requested):
It looks like login is broken. I'm not sure what's causing it, unfortunately.

mneedleman (dismissed):
🚢

frontend/src/FullPageRoutes/Matching.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }).catch((err) => console.error(`Failed to report match: ${err}`))
+    next()
+  }
+
+  return (
+    <Container>
+      {match ? (
+        <>
+          <Typography>
+            {match.first_name} {match.last_name}
```
rao53, line 72: Maybe put Email here so it's easier to see in the demo. I'm not sure but I'm guessing the edit profile and dashboard PRs might remove the need for that.

rao53 (dismissed):

frontend/src/FullPageRoutes/Matching.tsx
jcarrete5: We can just populate those fields in the database and we should be fine for the demo.

mneedleman (approved):

PR #47, opened 2021-02-02
Remove generation of and existing declaration files

jcarrete5:


jcarrete5:
This looks like a lot of changes because of the line ending normalization. Only the .gitattributes file was added for 32f17be

mneedleman (approved):

PR #48, opened 2021-02-03
Hooks

jcarrete5:
Originally I planned to have the pre-commit hook auto format the code, however, this proved to cause undesirable side effects and so the hook will only check if the code is properly formatted.

Problems formatting the code required stashing changes before formatting and popping afterwards. Popping would sometimes auto-merge code into the index which is undesirable. Also the idea of altering the code, regardless of the affect on functionality, without the review of the developer is undesirable, especially if black or isort does happen to cause an issue (unlikely but possible).

As it stands, the pre-commit hook will only check that python code is formatted correctly; Typescript formatting will be added later.


mneedleman (dismissed):

jcarrete5:
I had some mistakes in the previous hook scripts. I also threw in the flake8 check and fixed the issues that it found.

jcarrete5:
I'm going to make the typescript checks a separate PR

mneedleman (approved):

PR #49, opened 2021-02-04
Abstract modal

rao53:
Screen Shot 2021-02-03 at 8 18 05 PM
Screen Shot 2021-02-03 at 7 45 16 PM

I created both a notification modal which you can see in one of the images. I also created the abstract modal as requested in the issue request as seen above.


rao53:
Screen Shot 2021-02-03 at 8 23 07 PM
I also created like an abstract button for whenever we need to make any new buttons in the future. It felt more convenient for me than using the material UI button

frontend/src/components/AbstractModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+}))
+
+
+export default function AbstractModal(props) {
+    const [modal, setModal] = React.useState<modalType>(props);
+    const classes = useStyles()
+
+    React.useEffect(() => {
+        setModal(props);
+      }, [props]);
```
mneedleman: By passing props to `useState`, you've set the default value. I don't believe you need to call `useEffect` to set it again. Although passing a prop and then using it in state is very weird. I would have expected one or the other, but not both.

frontend/src/components/Button.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+    const { text, size, color, variant, onClick, ...other } = props
+    const classes = useStyles();
+
+    return (
+        <MuiButton
+            variant={variant || "contained"}
+            size={size || "large"}
+            color={color || "primary"}
+            onClick={onClick}
```
mneedleman: The alternative values here can just be the defaults. If you define some type here:
`const ButtonType = {
    variant?: variantType,
    size?: sizeType
    ....
}`.

Then you can set the defaults.
The ? next to the name marks the argument as optional.
Check out the Pick function for grabbing exact types from MUI. I think it would end up looking something like this:
`const ButtonType = Pick<MuiButtonType, "variant" | "size" | "color" | "onClick"> & {
...other stuff
}`

mneedleman (changes requested):
Really good work, just some small things that might help clean this up a bit

frontend/src/components/AbstractModal.tsx
```diff
@@ -0,0 +1,66 @@
+import { Dialog, DialogContent, DialogTitle, DialogActions, Typography, makeStyles, IconButton } from '@material-ui/core';
+import CloseIcon from '@material-ui/icons/Close';
+import * as React from 'react';
+import Button from './Button';
+
+
+export type modalType =  {
```
jcarrete5: As a convention, type names should be PascalCase
```suggestion
export type ModalType =  {
```

frontend/src/components/AbstractModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+    modalTitle: {
+        position: 'absolute',
+        top: theme.spacing(2),
+        right: theme.spacing(0.5)
+    }
+}))
+
+
+export default function AbstractModal(props) {
+    const [modal, setModal] = React.useState<modalType>(props);
```
jcarrete5:
```suggestion
    const [modal, setModal] = React.useState<ModalType>(props);
```

Also why are all of the props part of the state? It appears that the only property that can change is `isOpen`

frontend/src/components/Notification.tsx
```diff
@@ -0,0 +1,55 @@
+import { Snackbar, makeStyles } from '@material-ui/core'
+import * as React from 'react';
+import Alert from '@material-ui/lab/Alert';
+
+
+export type notificationType =  {
```
jcarrete5:
```suggestion
export type NotificationType =  {
```

frontend/src/components/Notification.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+
+const useStyles = makeStyles(theme => ({
+    root:{
+        top: theme.spacing(9) //Change this depending on size of header
+    }
+}));
+
+export default function Notification(props) {
+    const [notify, setNotify] = React.useState<notificationType>(props);
```
jcarrete5:
```suggestion
    const [notify, setNotify] = React.useState<NotificationType>(props);
```

frontend/src/components/AbstractModal.tsx
jcarrete5: From my understanding, this is updating the modal state every time the props change on a re-render. This step seems unnecessary especially because the only value that changes here is `isOpen` and I'd say it should be removed. Everything else should be referred to from the props object and removed from the state.

frontend/src/components/Button.tsx
```diff
@@ -0,0 +1,30 @@
...
+const useStyles = makeStyles(theme => ({
+    root: {
+        margin: theme.spacing(0.5)
+    },
+    label: {
+        textTransform: 'none'
+    }
+}))
+
+export default function Button(props) {
```
jcarrete5: Why do we need a whole separate component to set default values on a Button? Why not just set the values wanted directly on the button where it is used? If specifying all the options is repetitive, you can always put the options in an object and spread the object to provide the props.

frontend/src/components/Notification.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const classes = useStyles()
+
+    React.useEffect(() => {
+        setNotify(props);
+      }, [props]);
```
jcarrete5: I don't understand this pattern of mirroring the props in the state. Why are we doing this? What is the purpose?

jcarrete5 (changes requested):

frontend/src/components/AbstractModal.tsx
rao53: I've changed the the implementation by using the props like an object now.

frontend/src/components/Button.tsx
rao53: I just removed the button component overall.

frontend/src/components/AbstractModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+    modalTitle: {
+        position: 'absolute',
+        top: theme.spacing(2),
+        right: theme.spacing(0.5)
+    }
+}))
+
+
+export default function AbstractModal(props) {
+    const {modal, setModal} = props;
```
mneedleman: So since you're not doing anything other than getting the attributes of `props`, you can use destructuring:
This:
`export default function AbstractModdal(props) {
    const {attribute1, attribute2} = props
}`
Becomes:
`export default function AbstractModal({attribute1, attribute2}) {
}`

Also, you should make sure to keep the attribute names consistent, as it helps in the compilation process.

frontend/src/components/AbstractModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const classes = useStyles()
+
+    return (
+        <Dialog open={modal.isOpen}  classes={{paper: classes.modal}}>
+            <DialogTitle>
+                <IconButton className={classes.modalTitle} onClick={()=>setModal({...modal, isOpen: false})}>
```
mneedleman: In a way, this is just mimicking a cancel, right?

frontend/src/components/Notification.tsx
```diff
@@ -0,0 +1,51 @@
...
+export default function Notification(props) {
+    const {notify, setNotify} = props
+    const classes = useStyles()
+
+    const handleClose = (e) => { 
+        setNotify({ 
+            ...notify,
+            isOpen: false
+        })
+    }
```
mneedleman: This might be something the parent component would pass in, and probably as an optional prop

frontend/src/components/Notification.tsx
```diff
@@ -0,0 +1,51 @@
...
+            onClose={handleClose}
+            >
+                {notify.message}
+            </Alert>
+        </Snackbar>
+    )
+}
+
+
+{/* <Notification notify ={notify} setNotify={setNotify} /> */}
```
mneedleman: I don't think we need this

mneedleman (changes requested):

frontend/src/components/Notification.tsx
rao53: It is optional now

frontend/src/components/AbstractModal.tsx
rao53: yup

frontend/src/components/AbstractModal.tsx
rao53: changed now no worries.

frontend/src/components/AbstractModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+                <Typography variant="h6">
+                    {props.title}
+                </Typography>
+                <Typography variant="subtitle2">
+                    {props.description}
+                </Typography>  
+            </DialogContent>
+            <DialogActions className={classes.modalAction}>
+                <Button variant="contained" onClick={handleCancel} color="secondary">{props.cancelText || "NO"}</Button>
+                <Button variant="contained" onClick={handleConfirm} color="default">{props.confirmText || "YES"}</Button>
```
mneedleman: This is ok for now, but really we'd want to set that explicitly in the function arguments

mneedleman (approved):
This is looking super clean! Love it

frontend/src/components/AbstractModal.tsx
rao53: It technically still runs the function you give it. Right now, the handle confirm and handlecancel if you look into the function actually runs your own function if you provide it. Right now, the default is simply to close the modal so people won't have to do that themselves regardless.

frontend/src/components/AbstractModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const handleCancel = () => { 
+        props.setisOpen(false)
+        if(props.handleCancel != null){
+            props.handleCancel()
+        }
+    }
+
+    const handleConfirm = () => { 
+        if(props.handleConfirm != null){
+            props.handleCancel()
```
jcarrete5:
```suggestion
            props.handleConfirm()
```

frontend/src/components/AbstractModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+        right: theme.spacing(0.5)
+    }
+}))
+
+
+export default function AbstractModal(props: ModalType) {
+    const classes = useStyles()
+
+    const handleCancel = () => { 
+        props.setisOpen(false)
```
jcarrete5: Make this prop part of this component's state so the parent doesn't have to worry about managing it

frontend/src/components/Notification.tsx
```diff
@@ -0,0 +1,47 @@
...
+    root:{
+        top: theme.spacing(9) //Change this depending on size of header
+    }
+}));
+
+export default function Notification(props: NotificationType) {
+    const classes = useStyles()
+
+    const handleClose = (e) => { 
+        props.setisOpen(false)
```
jcarrete5: Make this prop part of this component's state so the parent doesn't have to worry about managing it

jcarrete5 (changes requested):

frontend/src/components/AbstractModal.tsx
rao53: If I do this, the parent would have to figure when to display the component and when not to display it. Thought it might be easier by just setting the state on the component itself.

mneedleman: I'd argue that the sole jobs of the modal are to display itself and to assign the callbacks (handleConfirm and handleClose). It does not need `setIsOpen`, presumably the calling component will do that in the two callbacks.

jcarrete5: @rao53 I feel I may have gave you conflicting advice the other day. I spoke with mike for clarification and agree that the parent should control when the modal is opened/closed. So essentially the `setisOpen` callback is redundant like mike mentions above because the other handlers will likely change some state that causes the modal to disappear.

jcarrete5 (changes requested):
Aside from the `setisOpen` comment, everything looks good to me.

frontend/src/components/AbstractModal.tsx
rao53: Removed the redundant isOpen prop.

frontend/src/components/Notification.tsx
```diff
@@ -0,0 +1,43 @@
...
+    root:{
+        top: theme.spacing(9) //Change this depending on size of header
+    }
+}));
+
+export default function Notification(props: NotificationType) {
+    const classes = useStyles()
+    
+    React.useEffect(() => { 
+    }, [props]);
```
mneedleman: Nope

frontend/src/components/AbstractModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+        right: theme.spacing(0.5)
+    }
+}))
+
+
+export default function AbstractModal(props: ModalType) {
+    const classes = useStyles()
+
+    React.useEffect(() => { 
+    }, [props]);
```
mneedleman: Nope

mneedleman (changes requested):

frontend/src/components/Notification.tsx
rao53: My mistake. I think I finally understand what you mean...

mneedleman (approved):

jcarrete5 (approved):

PR #51, opened 2021-02-04
Typecript hooks

jcarrete5:
This is a recreation PR #50 rebased on the hooks branch 

With this pull request, I am adding lint and format checks for typescript. I also enabled strict mode for stricter type checking. This uncovered a few potential errors in the code that I will be fixing in this PR. There are a lot of changes that are literally just reformatting code so I'd ask that in this PR you start by examining dffa903. In this commit, I had to make some decisions to conform to the stricter type checking rules that may not coincide with the original intent of the code. Specifically there where a couple `useState` calls that didn't correctly state if null was to be a valid value for the state variable.

Another important issue regards the [`http`](https://github.com/Conteo-Inc/conteo/blob/dffa90333295fff6d79d3163ec475f0ae09fe2d4/frontend/src/utils/fetch.tsx#L8-L10) helper function. The `parsedBody` attribute may be left undefined here and this is not checked for where it is referenced in other parts of the code (See dffa903).

This PR depends on PR #48 and thus should be considered/merged _after_ the PR #48.


mneedleman (dismissed):

rao53 (dismissed):
looks good

frontend/package.json
```diff
@@ -6,11 +6,18 @@
   "scripts": {
     "start": "webpack --mode development --watch",
     "build": "webpack --mode production",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "lint": "eslint src/",
+    "lint-check": "eslint --max-warnings 0",
+    "format": "prettier --write src/",
+    "format-check": "prettier --check"
```
mneedleman, line 13: Can you include an update to the readme with any common linting/formatting commands we should be using?

mneedleman (dismissed):
Readme updates aside, great work here

mneedleman (approved):
👍 🚢 :shipit: 👍

elp49 (approved):

PR #52, opened 2021-02-08
Authentication Overhaul

mneedleman:
1. No longer usering JWT. It got us off the ground,  but now it's causing issues.
2. I did what I could with `UserProfile` and `MatchStatus`. If it's non-breaking, please just fix this in your next related PR. If it is breaking, I need to know ASAP.

This is a major, blocking bug. I need review on this immediately.


frontend/src/utils/fetch.tsx
```diff
@@ -1,3 +1,7 @@
+//This is not ideal, but because of how the module is exported, it requires
+//this AFAIK - Michael
+import Cookies = require('js-cookie');
```
jcarrete5, line 3: Pretty sure this has to do with the module resolution setting in tsconfig. This should be fixed in #51

api/views.py
```diff
@@ -1,106 +1,63 @@
...
-        return Response(response, status=status_code)
+        logout(request=request)
 
+        return response.Response(status=status.HTTP_200_OK)
 
-class UserProfileView(RetrieveAPIView):
 
-    permission_classes = (IsAuthenticated,)
-    authentication_class = JSONWebTokenAuthentication
+class ProfileView(generics.RetrieveAPIView):
```
jcarrete5: This whole view is a bit weird. It is a retrieve view but nowhere is the serializer class or queryset being used. This is effectively a regular `APIView` and the following two lines (53 and 54) serve no purpose.

jcarrete5 (approved):
I made two comments but they aren't serious. Please look at them before merging

mneedleman:
foo

mneedleman:
bar

api/migrations/0003_profile.py
```diff
@@ -0,0 +1,53 @@
...
+                    "id",
+                    models.AutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("first_name", models.CharField(blank=True, max_length=50)),
+                ("last_name", models.CharField(blank=True, max_length=50)),
```
elp49, line 29: We will need to max length values to be retrievable by the frontend (unless we decide to hardcode it). Can we pull them out into a config?

jcarrete5: This is a good point. This is kind of awkward to do because we aren't building the form on the backend e.g. a template which would automatically grab this information from the model.

I'm not really sure of an elegant way to handle this atm with our current stack.

jcarrete5: Pulling it out into a data file like you mentioned might be workable. It can be fetched when the client navigates to the login page.

elp49: That sounds like what we want. Looking at the gender options, this could be a recurring issue in other areas as well

mneedleman: I think the duplication of the max lengths is not too egregious

jcarrete5: Having to edit the same data in two separate places is bound to cause consistency issues between the frontend and backend. And like eddie mentions, this goes beyond just these two fields. Any constraints on user-provided data on the backend should be reflected in the frontend.

jcarrete5: If we want frontend form validation, we need to coordinate this somehow. If we choose not to have frontend validation (at least not in the form of input element attributes), then I'd say it's safe to ignore this issue and let the backend say what is invalid and update the client accordingly (ofc we wouldn't want to be too specific with our error responses for logging in).

elp49: If we do decide to separate these conditions, then we could make use of it in models.py as well

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -21,14 +20,23 @@ const useStyles = makeStyles({
...
     const classes = useStyles();
 
     const [username, setUsername] = React.useState<string>(null);
     React.useEffect(() => {
-        request<User>('/api/current_user/', 'get', true, false).then(
-            (user) => {
-                setUsername(user.parsedBody.username);
+        request<UserProfile>('/api/profile/', 'get', true).then(
+            (profile) => {
+                setUsername(profile.parsedBody.first_name);
```
elp49, line 39: I'm considering changing UserProfile to this after I pull these changes. what do you think?

```
type UserProfile = {
    username: string
    data: {
        first_name: string
        last_name: string
        phone_number: string
        age: string | number
        gender: string
    }
}
```

elp49: and then removing the setUsername hook.

mneedleman: I feel a more elegant solution would be to use a better serialization strategy on the backend

api/migrations/0003_profile.py
mneedleman: See discord

frontend/src/FullPageRoutes/Profile.tsx
elp49: agreed

elp49 (approved):

jcarrete5 (approved):

PR #53, opened 2021-02-08
Adding editable profile inputs

elp49:
Capture
Capture2


frontend/src/components/ProfileContent.tsx
```diff
@@ -101,13 +173,39 @@ export default function ProfileContent(props: ProfileContentProps) {
...
+            }
+            { isEditMode ?
+                <Grid id={editableContent} container item alignItems="center" className={classes.fieldsContainer} xs={12}>
+                    <Grid id={editableContent} container item alignItems="center" xs={12}>
+                        <Grid item className={classes.field} sm={3} md={6}>
+                            <form onSubmit={saveFields}>
+                                <TextField required label={editedProfileFields.name.title} defaultValue={editedProfileFields.name.value} onChange={(e) => { setName(e.target.value) }} />
+                                <TextField disabled label={editedProfileFields.username.title} defaultValue={editedProfileFields.username.value} onChange={(e) => { setUsername(e.target.value) }} />
+                                <TextField required label={editedProfileFields.age.title} defaultValue={editedProfileFields.age.value} onChange={(e) => { setAge(e.target.value) }} />
+                                <TextField label={editedProfileFields.gender.title} defaultValue={editedProfileFields.gender.value} onChange={(e) => { setGender(e.target.value) }} />
```
mneedleman: Since the `onChange` is just passing the argument through, I believe you can just use:
`onChange={setThing}`

frontend/src/components/ProfileContent.tsx
```diff
@@ -50,46 +61,107 @@ const useStyles = makeStyles({
...
+    // original values.
+    const [name, setName] = React.useState<string>()
+    const [username, setUsername] = React.useState<string>()
+    const [age, setAge] = React.useState<number | string>()
+    const [gender, setGender] = React.useState<string>()
+    const [occupations, setOccupations] = React.useState<string[]>()
+    const [location, setLocation] = React.useState<string>()
+    const [interests, setInterests] = React.useState<string[]>()
+    const [religion, setReligion] = React.useState<string>()
+    const [editedProfileFields, setEditableProfileFields] = React.useState<ProfileFieldList | null>(null)
```
mneedleman: This seems like a good time to simplify this.

What do you think about creating a custom hook `useProfile`, which would return a profile object, and an update object. The update object could have attributes: `setName, setUsername,etc...`. In essence, the hook will just be comprised of multiple, smaller hooks

frontend/src/components/ProfileContent.tsx
```diff
@@ -50,46 +61,107 @@ const useStyles = makeStyles({
...
+
+    const readOnlyContent = 'readonly';
+    const editableContent = 'editable';
```
mneedleman: Why not just put these inline?

frontend/src/components/ProfileContent.tsx
```diff
@@ -101,13 +173,39 @@ export default function ProfileContent(props: ProfileContentProps) {
...
                 </Grid>
             </Grid>
-            <Grid container item alignItems="center" className={classes.fieldsContainer} xs={12}>
-                {profileFields.map(({ title, value }: ProfileField) => (
-                    <Grid key={title} item className={classes.field} sm={12} md={6}>
-                        <Typography>{title}: {value}</Typography>
+            { isEditMode ? null :
+                <Grid id={readOnlyContent} container item alignItems="center" className={classes.fieldsContainer} xs={12}>
+                    <Button variant="contained" color="secondary" className={classes.bottomRight} onClick={handleEditBtnClick}>Edit</Button>
+                    <Grid id={readOnlyContent} container item alignItems="center" xs={12}>
```
mneedleman: Two elements with the same id seems like it might cause problems

frontend/src/components/ProfileContent.tsx
```diff
@@ -50,46 +61,107 @@ const useStyles = makeStyles({
...
+        bottom: '0',
+        right: '0',
+        transform: 'translate(-10%, -10%)',
+    },
+    bottomRight2: {
+        position: 'absolute',
+        bottom: '0',
+        right: '0',
+        transform: 'translate(-100%, -10%)',
+    },
```
mneedleman: I'm a little concerned about using `position: absolute` anywhere except in the most extreme cases.

frontend/src/components/ProfileContent.tsx
```diff
@@ -50,46 +61,107 @@ const useStyles = makeStyles({
...
+            const { title, value } = editedField
+
+            // Test if field was changed.
+            if (originalField.value != value) {
+                console.log(`${title} was changed to: ${value}`)
+                originalField.value = value
+            }
+            else {
+                console.log(`${title} was not changed`)
+            }
```
mneedleman: There's a better way to do this, and I'm sure it exists somewhere in https://ramdajs.com/docs/, but I can't come up with it at the moment.

frontend/src/components/ProfileContent.tsx
```diff
@@ -101,13 +173,39 @@ export default function ProfileContent(props: ProfileContentProps) {
...
+                                    <Typography id={`${key}-readonlyFieldText`}>{title}: {value}</Typography>
+                                </Grid>
+                            );
+                        })}
                     </Grid>
-                ))}
-            </Grid>
+                </Grid>
+            }
+            { isEditMode ?
```
mneedleman: I feel like you could hide the first and show the second if isEditMode=true,
and show the first and hide the second otherwise.

This would get rid of the messy `: null` result here.

mneedleman (changes requested):
Great work! I do have some small suggestions though

frontend/src/components/ProfileContent.tsx
elp49: I thought so too, and maybe there is a clever way to do this, but it throws:
TS2322: Type 'Dispatch<SetStateAction<string>>' is not assignable to type '(event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => void'.

frontend/src/components/ProfileContent.tsx
elp49: yes, this was a copy and paste error

frontend/src/components/ProfileContent.tsx
elp49: I find it better practice to pull IDs out into a const to avoid typos and it makes it easier to change the ID in the future

frontend/src/components/ProfileContent.tsx
elp49: it's a powerful CSS style. as long as the correct parent has `position: relative` set, it simplifies the jsx

frontend/src/components/ProfileContent.tsx
elp49: yes, this is still a draft. I havent decided how it should work yet

frontend/src/components/ProfileContent.tsx
elp49: I used to know of a way to hide something like this easily but I cant find it online anywhere. the closest thing I found to do this without separating them into components is `visibility: hidden`, but that still renders the jsx and leaves room for it but just makes it invisible which is not what we want.

I'll look into better options

frontend/src/components/ProfileContent.tsx
mneedleman: ahhh

frontend/src/components/ProfileContent.tsx
mneedleman: Does `{isEditMode && etc}` work?

mneedleman: Or maybe just one &?

elp49: ahh yes, that worked great

mneedleman:
Can you add screenshots as well?

frontend/src/components/ProfileContent.tsx
```diff
@@ -50,64 +61,199 @@ const useStyles = makeStyles({
...
+    }
 
-    // Add user profile values to list.
-    var profileFields: ProfileField[] = [
-        {
+    // Pass props to to useProfile hook.
+    const { editableContent, setters } = useProfile(props)
+
+    // Initialize readonly profile content and acquire hook to update it when edits are saved.
+    const [readonlyContent, setProfile] = React.useState<ProfileContentProps>(editableContent);
```
elp49: @Deravadom is this what you had in mind? it seems to work very well

mneedleman: Exactly

mneedleman: Although I'd personally pull it out of `src/components/ProfileContent.tsx` and into `src/utils/profile.tsx` or something similar.

frontend/src/components/ProfileContent.tsx
```diff
@@ -1,29 +1,41 @@
...
-    religion: string;
-    location: string;
-    occupations: string[];
-    age: number | string,
-    interests: string[];
-};
+    name: string
+    username: string
+    age: number
+    gender: string
```
AyalaCnaan: I don't know how to change this, but I think that 'string' is too loose a type here. We only have three string options: Male, Female, and Other. We can't write 'Bozo' for instance and get anywhere.

frontend/src/components/ProfileContent.tsx
AyalaCnaan: I'm also a little leery of using 'absolute.' It's not the hill I'm going to do Tai Chi on, but it is something to pay attention to long term.

frontend/src/components/ProfileContent.tsx
```diff
@@ -50,64 +62,160 @@ const useStyles = makeStyles({
...
+            value: readonlyContent.location,
         },
-        {
+        interests: {
             title: 'Interests',
-            value: props.interests.join(', ')
+            value: readonlyContent.interests.join(', '),
         },
-        {
+        religion: {
```
AyalaCnaan: It strikes me that if people can have multiple occupations, they can have multiple religions. Just a thought. It's perfectly compatible to be Jewish and Atheist. Or Unitarian and Buddhist.

AyalaCnaan (dismissed):
Eventually, we need to shift age from a number to some sort of date value because people get older and don't want to have to change their age every year. 
Screenshots look good in general. At some point we need to have an annoying conversation about subtle background differences. But, that can wait until we're in CSS polish mode.

frontend/src/components/ProfileContent.tsx
jcarrete5: I'm not really too concerned about the use of absolute here. If it becomes problematic, then it can be changed later when we focus more on styling the website.

jcarrete5 (dismissed):

frontend/src/components/ProfileContent.tsx
jcarrete5: In this instance, I think using [`e.currentTarget.value` is more accurate than `e.target.value`](https://developer.mozilla.org/en-US/docs/Web/API/Event/Comparison_of_Event_Targets). In general, this is usually the case because currentTarget will always be the element this listener is attached to.

frontend/src/components/ProfileContent.tsx
mneedleman: Absolute messes with everything else on the page. How does it look on mobile, for example?

jcarrete5: I mean it looks about as good as everything else. I just don't think its worth fiddling with the exact position of things when we are just trying to get things to work.

image

frontend/src/components/ProfileContent.tsx
sb3958: Isn't the string type, the type it should be when it is pulled from the database? Also, although it is set as a string, won't the user have the options to change from either Male, Female or Other without having to type in any information?

AyalaCnaan: I don't know TypeScript well enough to say what is should be, but possibly an enum. Leaving it as just a string is too loose.

frontend/src/components/ProfileContent.tsx
elp49: agreed

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -19,64 +32,80 @@ const useStyles = makeStyles({
...
-      if (profile.parsedBody) {
-        setUsername(profile.parsedBody.first_name)
-      } else {
-        console.error(
-          "FIXME: json.parsedBody was undefined and so the login failed. \
-                  This is a problem with the code that needs to be addressed"
-        )
-      }
-    })
+    request<UserProfile>("/api/current_user/", "get", true, false)
```
mneedleman: That endpoint no longer exists.

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -1,10 +1,23 @@
...
+type UserProfile = {
+  username: string
+  data: {
+    first_name: string
+    last_name: string
+    phone_number: string
+    age: string | number
+    gender: string
+  }
+}
```
mneedleman: I'd rather these be unnested. Instead, the serializer should combine data into an object with primitives, not an object with another object. Also age should be a number

frontend/src/components/ProfileContent.tsx
```diff
@@ -51,78 +75,259 @@ const useStyles = makeStyles({
...
+                  md={6}
+                >
+                  <Grid item xs={9}>
+                    <Typography>
+                      {title}: {value}
+                    </Typography>
+                  </Grid>
+                </Grid>
+              ))}
+              <Grid container item justify="center" xs={12}>
```
mneedleman, line 194: I believe this throws an error if a Grid is both a container and an item

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -19,64 +32,80 @@ const useStyles = makeStyles({
...
+  const dummyUserProfile: UserProfile = {
+    username: "nope",
+    data: {
+      first_name: "nope",
+      last_name: "nope",
+      phone_number: "nope",
+      age: -1,
+      gender: "nope",
+    },
+  }
```
mneedleman: It seems odd to create what is essentially default values, then immediately attempt to overwrite them. If the request fails, the user will see "nope" everywhere, which is probably not what we want.

frontend/src/components/ProfileContent.tsx
```diff
@@ -51,78 +75,259 @@ const useStyles = makeStyles({
...
+                  <Grid item xs={9}>
+                    <TextField
+                      label={title}
+                      defaultValue={value}
+                      {...textFieldProps}
+                    />
+                  </Grid>
+                </Grid>
+              ))}
+              <Grid container item justify="center" xs={12}>
```
mneedleman: Again, a grid should not be both a container and an item

mneedleman (changes requested):
Getting there! There are just some small issues I'd like to see addressed

frontend/src/components/ProfileContent.tsx
jcarrete5: This is fine: https://material-ui.com/components/grid/#nested-grid

mneedleman: Weird, I got warnings when I tried it

jcarrete5: What are the warnings? I don't have warnings.

mneedleman: I'll look into another time

frontend/src/FullPageRoutes/Profile.tsx
elp49: this was just for testing

frontend/src/components/ProfileContent.tsx
```diff
@@ -51,78 +68,198 @@ const useStyles = makeStyles({
     padding: 5,
     paddingRight: 10,
   },
+  bottomRight: {
+    position: "absolute",
+    bottom: "0",
+    right: "0",
+    transform: "translate(-10%, -10%)",
```
mneedleman, line 75: I don't like `position: "absolute"` at all. But this is fine, for now

mneedleman (approved):

PR #55, opened 2021-02-12
User delete request on backend

rao53:
This backend call deletes both the user and profile object. However, a lot of objects depend on the user object. So, what should the actual definition of deleting be. Should we delete only the user object by making every field in it null except the primary key.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -29,16 +29,9 @@ class UserRegistrationView(generics.CreateAPIView):
 class UserAccountDeleteView(views.APIView):
     permission_classes = (permissions.AllowAny,)
     def delete(self, request):
-        try:
-            request_id = request.data["id"] 
-            user = User.objects.get(pk=request_id)
-            profile = Profile.objects.filter(user=user)
-            profile.delete()
-            user.delete()
-            return response.Response(status=status.HTTP_200_OK)
-        except User.DoesNotExist:
-            return response.Response(status=status.HTTP_400_BAD_REQUEST)
-        return response.Response(status=status.HTTP_204_NO_CONTENT)
+        req_user = request.user
+        req_user.delete()
+        return response.Response(status=status.HTTP_200_OK)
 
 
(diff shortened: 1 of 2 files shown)


mneedleman:
We should have a meeting and discuss how videos should be handled

api/views.py
```diff
@@ -26,6 +26,20 @@ def post(self, request: request.Request):
...
 
+class UserAccountDeleteView(views.APIView):
+    permission_classes = (permissions.AllowAny,)
+    def delete(self, request):
+        try:
+            request_id = request.data["id"] 
+            user = User.objects.get(pk=request_id)
+            profile = Profile.objects.filter(user=user)
+            profile.delete()
+            user.delete()
```
mneedleman: So we should get the requesting user from the request itself, not the data:
`user = request.user` 
This will eliminate the need for the try...catch
Also it should be possible to just delete the user, and let the profile delete through the cascade.

mneedleman (changes requested):
Also I'd love to see a test case for this

api/views.py
rao53: so, you're saying the request is the user object itself, not the id of the user?

rao53 (commented):
> Also I'd love to see a test case for this

Using a factory right? Where would I put this test case. I was hoping that we could hold off on this until your factory classes are pushed to master...

api/views.py
mneedleman: Not quite, but request.user works just fine as the user

rao53: It works fine you're right

mneedleman:
> 
> 
> > Also I'd love to see a test case for this
> 
> Using a factory right? Where would I put this test case. I was hoping that we could hold off on this until your factory classes are pushed to master...

Good point haha

elp49 (approved):
Good job

mneedleman (approved):

PR #56, opened 2021-02-14
Reporting backend

jcarrete5:


api/models.py
```diff
@@ -57,3 +57,42 @@ class Meta:
...
+
+class Report(models.Model):
+    class Status(models.TextChoices):
+        UNASSIGNED = "U"
+        ASSIGNED = "A"
+        RESOLVED = "R"
+
+    class ReportType(models.TextChoices):
+        VIDEO = "V"
+        PROFILE = "P"
```
mneedleman, line 70: Interesting way of doing this. Is there an advantage over an enum?

api/models.py
```diff
@@ -57,3 +57,42 @@ class Meta:
...
+
+    status = models.CharField(
+        max_length=1, choices=Status.choices, default=Status.UNASSIGNED
+    )
+    report_type = models.CharField(max_length=1, choices=ReportType.choices)
```
mneedleman, line 75: These can be ChoiceFields

api/models.py
```diff
@@ -57,3 +57,42 @@ class Meta:
...
+    description = models.TextField(blank=True)
+    reporter = models.ForeignKey(
+        User,
+        on_delete=models.SET_NULL,
```
mneedleman, line 79: Good idea

mneedleman (changes requested):
Overall very clean and straightforward! The only real change I'd want to see is to make those Charfields be ChoiceFields.

api/models.py
jcarrete5: As far as I know, ChoiceField is not a thing for models. It only exists for form fields. To limit choices on a model, you use the choices kwarg.

api/models.py
jcarrete5: These are basically enums provided by django that are designed to be easy to use as choices for model fields. https://docs.djangoproject.com/en/3.1/ref/models/fields/#enumeration-types

api/serializers.py
```diff
@@ -38,3 +38,9 @@ class VideoSerializer(serializers.ModelSerializer):
     class Meta:
         model = Video
         fields = ("id", "title")
+
+
+class ReportSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Report
+        fields = ("report_type", "reporter", "reportee", "description")
```
elp49, line 46: why not include submitted_on and status?

elp49 (approved):
Looks good

api/serializers.py
jcarrete5: As of right now, I just don't know what data will be wanted on the frontend so I just left it as these. I didn't want to use `__all__` because I favor being explicit in the serializer in case fields are added to the model that we don't want to serialize.

api/models.py
mneedleman: Oops, my mistake

mneedleman (approved):

PR #57, opened 2021-02-15
Help and Contact Us Page

sb3958:
Contact Page
Contact Us Page

Help Page
Help Page


elp49:
@Deravadom should the Help and Contact Us be components or full pages? and if components, then which pages will use them or will the dashboard bring them up in modules?

frontend/src/components/Help.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    </Grid>
+                    <Grid item>
+                        <Card className={classes.root}>
+                            <CardContent>
+                                <CardActions>
+                                <Button href="#ContactUs"> Support Center</Button>
+                                </CardActions>
+                            </CardContent>
+                        </Card>
+                    </Grid>
```
elp49: this isn't super urgent but we could automate this jsx by pulling the button values out into a const like:

`const helpBtns = [
  {
    href="ContactUs"
    value="Contact Us"
  },
  ...
]`

elp49 (changes requested):
Good job! I have a few small changes requests and then it should be good to go

frontend/src/components/ContactUs.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const [values, setValues] = useState(contactValue);
+    const classes = useStyles();
+
+    const getInput= e =>{
+        const {name, value } = e.target
+        setValues({
+            ... values,
+            [name]: value
+        })
+    }
```
elp49: This is a clever way of setting the values we need. I didn't think of doing it this way, but it's nice in that it requires less code than what I've done in the [Profile Editing Tool](https://github.com/Conteo-Inc/conteo/pull/53/files). @Deravadom and I decided to create a hook for each value that needs to be set and move them to frontend/src/utils/profile.tsx 

However we decide to set values in a form, it should be consistent throughout the app (unless there's a good reason to make an exception). If we decide to do it this way, then we definitely should pull out the `name` attribute into a const along with `label` and `value` like I mentioned in Help.tsx

frontend/src/components/ContactUs.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+                <Paper className={classes.pageContent}>
+                    <form className={classes.root}>
+                        <Grid container spacing={2} direction="column" justify="center">
+                            <Grid item xs={12}>
+                                <TextField
+                                variant="outlined"
+                                label="Your name"
+                                name = "name"
+                                value={values.name}
```
elp49: What exactly does setting `value` do here? I think I remember having issues in the Profile Editing branch when I set the value attribute because it wouldn't let me edit the value once this attribute is set.

sb3958: Setting the value here is to get the information from the user.  The information then would be used to update the contact value when the getInput function is called.

mneedleman:
> 
> 
> @Deravadom should the Help and Contact Us be components or full pages? and if components, then which pages will use them or will the dashboard bring them up in modules?

I think these are fine as full pages. It might be worth putting together a common form component. But for now this is fine

frontend/src/components/ContactUs.tsx
mneedleman: I think a combination of the two might be beneficial. I like the choice to wrap the values in the `contactValue` object. That's something I'd like to see in the profile editing tool. Also the name here is a bit misleading. `getInput` actually sets the values.

frontend/src/components/ContactUs.tsx
```diff
@@ -0,0 +1,165 @@
...
+            ... values,
+            [name]: value
+        })
+    }
+
+    return (
+        <>
+       
+        <Grid container>
+            <Paper className={classes.pageSideBar}>
```
mneedleman: Any reason you didn't just apply the class to the Grid? What does Paper add? I've never used it before.

frontend/src/components/Help.tsx
```diff
@@ -0,0 +1,85 @@
+import * as React from 'react';
+import { TextField, Grid, makeStyles, Button, Typography, Card, CardContent, CardActions, Paper } from '@material-ui/core'
+
+
+const useStyles= makeStyles({
+    root:{
+        
+        
+    },
```
mneedleman: No need for an empty class here

mneedleman (changes requested):
There are a lot of spacing issues here that suggest you need to run this through a formatter. Also you need to set up the pre-commit hooks.

mneedleman:
I'm a little confused here. The task asks for a Help page and an FAQ section. I don't see an FAQ section here. Also it doesn't specify a Contact Us form, since that's part of a separate task. And the form shown says "Send Report".

frontend/src/components/ContactUs.tsx
sb3958: Paper just applies a background that resembles a flat paper.  So I just put the grid inside the paper.  I can try apply the class to the Grid.

AyalaCnaan:
> @Deravadom should the Help and Contact Us be components or full pages? and if components, then which pages will use them or will the dashboard bring them up in modules?

While I am not @Deravadom I will say that in my opinion it should be full pages. That seems to be the industry standard. This is not, however, deeply important to me if someone has a strong view to the contrary.

frontend/src/components/ContactUs.tsx
AyalaCnaan: Perhaps change getInput to setInput.

AyalaCnaan:
This looks like it's moving along nicely. The FAQ section needs to be addressed or moved to its own task. The style is good.

mneedleman (changes requested):
Did you test these changes before pushing? You cannot have multiple default exports in a module. Also you defined `ContactUs` twice.

There are still severe spacing issues here. You need to be using the pre-commit hooks and the npm scripts. If you need help, ask one of us to show you.

frontend/src/FullPageRoutes/Help.tsx
```diff
@@ -0,0 +1,96 @@
...
+      href: "ContactUs",
+      value: "Contact Us",
+    },
+    {
+      href: "FAQs",
+      value: "FAQs",
+    },
+    {
+      href: "Tutorials",
+      value: "Tuturials",
```
mneedleman: Small typo here

mneedleman (dismissed):
We can address small changes at a later date

frontend/src/App.tsx
```diff
@@ -86,6 +90,12 @@ export default function App(): JSX.Element {
...
             </Route>
+            <Route path="/Help">
+              <Help />
+            </Route>
+            <Route path="/ContactUs">
+              <ContactUs />
+            </Route>
             <Route path="/Profile">
               <ProfilePage />
             </Route>
```
elp49: These should be ProtectedRoute components

elp49 (approved):

frontend/src/FullPageRoutes/Help.tsx
elp49: fixed

mneedleman (approved):

PR #59, opened 2021-02-25
Video viewing

mneedleman:
This ended up being a lot lighter than expected. Also I fixed the errors on login and videolist.


rao53 (dismissed):
Everything looks good based on the tests you wrote. I can't really test it on the frontend yet. I was also hoping you could clarify the difference between the videoView and videoList page. I'm not necessarily sure what the difference is.  The login bug is fixed too.. I checked that. Apart from that, solid job in all..

mneedleman:
The VideoListPage is very much obsolete at this point. It served originally as a test for video viewing, and laid the groundwork for what the dashboard will ultimately become. But now it is no longer necessary.

rao53:
> The VideoListPage is very much obsolete at this point. It served originally as a test for video viewing, and laid the groundwork for what the dashboard will ultimately become. But now it is no longer necessary.

You should probably delete it...

rao53 (approved):

frontend/src/FullPageRoutes/VideoViewPage.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export default function VideoViewPage(): JSX.Element {
+  // In this case, focusedUser will be the one whose video I am viewing
+  const [focusedUser] = useFocusedUser()
+  const [video, setVideo] = React.useState<VideoType>({
+    id: 0,
+    receiver: 0,
+    video_file: "",
+  })
+
```
AyalaCnaan, line 21: Can you explain to me why you don't just set the values in the type VideoType?

AyalaCnaan (approved):
Looks good to me.

frontend/src/FullPageRoutes/VideoViewPage.tsx
mneedleman: Because `VideoType` is a type declaration, it can't hold values itself. But the variable `video`, which is of type `VideoType`, holds the values

PR #60, opened 2021-02-25
Profile Intro Video

mneedleman:
Wrapped in this is a duplicated update to the AppContext. Should be fine once merge conflicts are resolved.

With video:
image

Without video:
image


frontend/src/components/ProfileContent.tsx
```diff
@@ -51,6 +53,13 @@ const useStyles = makeStyles({
     padding: 5,
     paddingRight: 10,
   },
+  recordButton: {
+    backgroundColor: Colors.DEEP_BLUE,
+    color: "white",
+    "&:hover": {
+      backgroundColor: Colors.DEEP_RED,
+    },
+  },
```
AyalaCnaan, line 91: This is in keeping with what we already have by way of buttons? Or what we plan to have?

AyalaCnaan (dismissed):
The only thing I'm not really understanding is where the video goes after it is recorded. But that's probably a me problem, not a code problem.

frontend/src/components/ProfileContent.tsx
mneedleman: We have a task planned to make the css coloring system consistent, using themes. For now, this is fine as long as it works

api/serializers.py
```diff
@@ -1,12 +1,26 @@
...
+        # get video
+        video = None
+        try:
+            video = Video.objects.get(
+                Q(sender=instance.user) & Q(receiver=instance.user)
+            )
+            video = read_video(video.video_file)
+        finally:
+            rep["video"] = video
+            return rep
```
elp49: does this need an `except` ?

frontend/src/components/ProfileContent.tsx
```diff
@@ -165,8 +175,32 @@ export default function ProfileContent({
...
+                src={readonlyContent.video}
+                width={"100%"}
+                height={"100%"}
+              />
+            ) : (
+              <Button
+                className={classes.recordButton}
+                size="large"
+                component={Link}
+                to="/record"
```
elp49: will the record page redirect back to profile?

elp49 (dismissed):

api/serializers.py
mneedleman: All I would do in the `except` is ensure the value of `video` is none. Didn't seem worth it, since `video` is initialized to none

frontend/src/components/ProfileContent.tsx
mneedleman: I could add that in

api/serializers.py
jcarrete5: I don't really get the try-finally here. Are you anticipating an exception in that code?

api/serializers.py
```diff
@@ -48,6 +62,18 @@ class Meta:
...
+def read_video(video):
+    """
+    @param video: A FieldFile representing the video.
+    Commonly found on Video.video_file
+    """
+    video.open()
+    try:
+        return video.read().decode()
+    finally:
+        video.close()
```
jcarrete5:
```suggestion
    try:
        return video.read().decode()
    finally:
        video.close()
```
You shouldn't need to explicitly open the file because [Django will open it for you](https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.fields.files.FieldFile.open).

rao53:
> The only thing I'm not really understanding is where the video goes after it is recorded. But that's probably a me problem, not a code problem.

Following up on this too. Not sure if this stored anywhere because I don't see anything on the database....

jcarrete5:
> > The only thing I'm not really understanding is where the video goes after it is recorded. But that's probably a me problem, not a code problem.
> 
> Following up on this too. Not sure if this stored anywhere because I don't see anything on the database....

The video file is saved when the Video model instance is saved in the VideoSerializer. It is put under `MEDIA_ROOT` defined in settings.py

rao53 (dismissed):

api/serializers.py
mneedleman: Yes, the get will fail if there is no video

PR #64, opened 2021-03-01
Profile Interests

mneedleman:
NOTE: This is branched off of #53 , and should NOT be merged until that one is.

image


sb3958:
Looks good. Like the format of it

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -19,58 +32,82 @@ const useStyles = makeStyles({
...
+          birthday: new Date(profile.birth_date),
+          gender: profile.gender,
+          interests: profile.interests,
+        }
+        setProfile(content)
+      })
+      .catch((err) => {
+        console.log(err)
+      })
+  }, [])
```
elp49: should we add a dependency here like readonlyContent?

elp49 (dismissed):

frontend/src/FullPageRoutes/Profile.tsx
mneedleman: I keep running into a disparity between what React documentation says and what I see in practice. The bug where render is called constantly is supposedly fixed by adding dependencies or the empty `[]`. However, I found that the bug was present with the two dependencies and not present with the empty `[]`.

PR #65, opened 2021-03-04
Routing and Context Overhaul Redo

mneedleman:
- Pulled AppContext.Provider in to a separate component
- Routes that should not be accessible by unauthorized users will now redirect to login
- PAGE REFRESH WORKS 💯 
- DIRECT ROUTING WORKS 💯 💯 💯 💯 
- Separated App into App, AppHeader, and AppFooter
- Added `/user` endpoint to check for the current user
- Added a loading page that displays while the system is checking if the user exists
- Pulled out login, register, and logout into independent functions with callbacks for routing or cleanup
- Added `User` to AppContext


frontend/src/utils/context.tsx
```diff
@@ -18,3 +40,73 @@ export default function useFocusedUser(): AppContextType["focusedUser"] {
   const context = React.useContext(AppContext)
   return context.focusedUser
 }
+
+type UseUserReturnType = {
+  user: Nullable<User>
+  logged_in: Nullable<boolean>
+  register: (authInfo: AuthInformation, callBack: CallBack) => void
+  login: (authInfo: AuthInformation, callBack: CallBack) => void
+  logout: (CallBack: CallBack) => void
```
jcarrete5: Instead of passing a callback to these functions, it would be nicer to return the promise associated with the request. As it stands right now, there is no error handling in these requests and no way to do it from the caller.

jcarrete5 (changes requested):
So we make a request in the context to set if the user is logged in. Sounds good. I don't see where focused user is recovered which might be undesirable.

Other than that, I like the useUser hook. We'll probably want more fields in the user type for the app than just email and first name later.

mneedleman:
Not sure what you mean by recovered, but it's by design that `focusedUser` is null here. What value would it hold here? It should only the user if they're interacting with profile (and possibly account), but otherwise I think null is fine.

frontend/src/utils/context.tsx
mneedleman: Interesting idea, I'll explore that

mneedleman: As far as I can tell, the two options for a return type are `Promise<void>` and `Promise<void | HttpResponse<Response>>>` Neither are great, but I'll settle for the former since we're not sending data back

jcarrete5:
> Not sure what you mean by recovered, but it's by design that `focusedUser` is null here. What value would it hold here? It should only the user if they're interacting with profile (and possibly account), but otherwise I think null is fine.

If the user was in the middle of doing something and focusedUser is set, then focusedUser will be null after a refresh because it is not recovered from somewhere. If this is by design, then whatever. I'm just pointing out that I didn't see it being recovered in case that was overlooked.

mneedleman:
Ohhhh, I see what you mean

frontend/src/utils/context.tsx
jcarrete5: Those functions don't return anything for the final `then` so `Promise<void>` is fine.

jcarrete5 (approved):

frontend/src/components/ProtectedRoute.tsx
```diff
@@ -0,0 +1,28 @@
...
+export function ProtectedRoute({
+  children,
+  ...rest
+}: ProtectedRouteProps): JSX.Element {
+  const {
+    user: [userValue],
+  } = React.useContext(AppContext)
+  const location = useStatefulLocation()
+
+  return (
```
rao53, line 17: This might be asking for too much but we are currently still able to go to the sign in/sign up page when logged in. It would be probably be weird to be able to log in if already logged in... Although, I think what we have now is sufficient enough for the presentation.

rao53 (approved):

frontend/src/components/ProtectedRoute.tsx
mneedleman: How did you do that?

rao53: After I was logged in, I simply changed the route back to /tokens and it took me back to the login/sign up page

mneedleman: Ah, I think that's fine

PR #69, opened 2021-03-07
site_flow_and_css

elp49:
Capture3

…from viewvideo page. adding sender id to video page params


rao53:
I think this task might be conflicting with the expanded video view one. The expanded video view currently allows for a user to view a received video but not to send a video. I believe we're also not using a view video button because we are using the mail icon to view received videos.

elp49:
> I think this task might be conflicting with the expanded video view one. The expanded video view currently allows for a user to view a received video but not to send a video. I believe we're also not using a view video button because we are using the mail icon to view received videos.

I'm not sure about this. the clickup task said to replace the dropdown icon with a view video button

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -5,6 +5,7 @@ import Controls from "../components/video/Controls"
 import { parseIdentity, request } from "../utils/fetch"
 import { useHistory, useParams } from "react-router-dom"
 import { Nullable, NullableId } from "../utils/context"
+import ConfirmationModal from "../components/AbstractModal"
```
mneedleman, line 8: As far as I can tell, there is no `ConfirmationModal` exported from AbstractModal.

frontend/src/components/AbstractModal.tsx
```diff
@@ -42,34 +42,48 @@ const useStyles = makeStyles((theme) => ({
...
+  confirmText,
+  cancelText,
+  isOpen,
+  setisOpen,
+  handleConfirm,
+  handleCancel,
+}: ModalType): JSX.Element {
   const classes = useStyles()
 
+  const handleAction = (action: any) => {
```
mneedleman: No any. Also are you using the pre-commit hooks? It should have yelled at you for this

frontend/src/components/AbstractModal.tsx
```diff
@@ -42,34 +42,48 @@ const useStyles = makeStyles((theme) => ({
...
+    setisOpen(false)
+    action()
+  }
+
   return (
-    <Dialog open={props.isOpen} classes={{ paper: classes.modal }}>
+    <Dialog open={isOpen} classes={{ paper: classes.modal }}>
       <DialogTitle>
-        <IconButton className={classes.modalTitle} onClick={props.handleCancel}>
+        <IconButton className={classes.modalTitle} onClick={() => {handleAction(handleCancel)}}>
```
mneedleman: There's no need to make the argument an anonymous function here

mneedleman (changes requested):
What's going on with your package-lock? Also please update the title. I know it's been confusing because of task duplication errors

frontend/src/FullPageRoutes/RecordPage.tsx
elp49: This imports the default function AbstractModal and I'm renaming it here to reflect it's implementation. I can rename it if you'd like

frontend/src/components/AbstractModal.tsx
elp49: This is so that the AbstractModal handles closing itself upon any action

frontend/src/FullPageRoutes/RecordPage.tsx
mneedleman: Oh weird. I didn't know you could rename the default like that

frontend/src/components/AbstractModal.tsx
```diff
@@ -42,34 +40,34 @@ const useStyles = makeStyles((theme) => ({
...
       </DialogContent>
       <DialogActions className={classes.modalAction}>
-        <Button
-          variant="contained"
-          onClick={props.handleCancel}
-          color="secondary"
-        >
-          {props.cancelText || "NO"}
+        <Button variant="contained" onClick={handleCancel} color="secondary">
+          {cancelText || "NO"}
```
mneedleman, line 67: If I'm being picky, this could be the default value of `cancelText`, since the parameter is optional

mneedleman (dismissed):
Nitpicking aside, looks good

frontend/src/components/AbstractModal.tsx
elp49: yeah you're right

mneedleman (approved):

AyalaCnaan (approved):
Looks good to me, but I'm not in the nitty gritty like you guys.

PR #76, opened 2021-03-09
Profile privacy settings

elp49:
8Capture
9Capture
1Capture
2Capture
3Capture
4Capture
5Capture
6Capture
7Capture


jcarrete5:
~~Don't delete the `package-lock.json` file~~

jcarrete5:
> Don't delete the `package-lock.json` file

nevermind. The diff looked weird but the file still exists

frontend/src/components/profile/ProfileSidebar.tsx
```diff
@@ -0,0 +1,154 @@
...
+import LockIcon from "@material-ui/icons/Lock"
+
+// ProfileSidebar props.
+type ProfileSidebarProps = {
+  firstName: string
+  lastName: string
+  componentStateSetters: ProfileComponentSetters
+}
+
+// Aitem listed in the sidebar.
```
jcarrete5:
```suggestion
// An item listed in the sidebar.
```

api/models.py
```diff
@@ -12,7 +12,43 @@ class Profile(models.Model):
...
+    )
+    interests_privacy = models.CharField(
+        max_length=2, choices=Setting.choices, default=Setting.PUBLIC
+    )
+
+
+class Interest(models.Model):
+    profile = models.ForeignKey(
+        Profile, on_delete=models.CASCADE, verbose_name="related profile"
+    )
```
jcarrete5: As we discussed the other day, this is probably better suited as a many-to-many field in the Profile.

api/views.py
```diff
@@ -82,20 +84,112 @@ def post(self, request):
...
+        try:
+            # Updated user interests.
+            updatedInterests = request.data.pop("interests")
+            self.updateInterests(updatedInterests, request.user.profile)
+        except KeyError:
+            # User did not updates their interests.
+            pass
+
+        # Update profile content.
+        return self.update(request=request)
```
jcarrete5: This is related to the previous comment and to what we discussed the other day. This would be simpler by taking advantage of a many-to-many field on the profile model.

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -33,93 +34,145 @@ const useStyles = makeStyles({
...
+  let birthday = null
+  if (birth_date !== null) {
+    birthday = new Date(birth_date)
+    // Typescript removes a day.
+    birthday.setDate(birthday.getDate() + 1)
+  }
+
+  return birthday
+}
+
```
jcarrete5: I'm a bit confused why we need to add 1 to the date here. Both the javascript Date and the python datetime.date appear to handle this correctly i.e. the date starts at 1. What was the problem you were having by not adding 1?

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -33,93 +34,145 @@ const useStyles = makeStyles({
...
+  }, [])
+
+  // Component state hooks.
+  const { componentStates, componentSetters } = useProfileComponents()
+  const {
+    isProfileActive,
+    isPrivacyActive,
+    // isNotificationsActive,
+    // isSettingsActive,
+    // isContactUsActive,
```
jcarrete5:
```suggestion
```

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -33,93 +34,145 @@ const useStyles = makeStyles({
...
+            {isNotificationsActive && (
+              // Add notifications component here
+            )}
+            {isSettingsActive && (
+              // Add settings component here
+            )}
+            {isContactUsActive && (
+              // Add contact us component here
+            )}
+          */}
```
jcarrete5:
```suggestion
```

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -0,0 +1,372 @@
...
+    height: 200,
+    width: 200,
+    display: "flex",
+    overflow: "hidden",
+    alignItems: "center",
+    borderRadius: "50%",
+    justifyContent: "center",
+  },
+  image: {
+    // color: "transparent",
```
jcarrete5:
```suggestion
```

frontend/tsconfig.json
```diff
@@ -8,7 +8,9 @@
         "noImplicitThis": true,
         "isolatedModules": true,
         "allowJs": true,
-        "moduleResolution": "node"
+        "moduleResolution": "node",
+        "resolveJsonModule": true,
+        "esModuleInterop": true
```
jcarrete5: Why was this added? To import json files in the code? Modifications to the build are almost always more appropriate in the webpack config since we are using webpack. If you need to import json (which shouldn't be necessary if the interests are stored on the backend), then a json-loader in the webpack config would be more appropriate.

jcarrete5 (changes requested):

api/models.py
mneedleman: I agree

frontend/src/FullPageRoutes/Profile.tsx
mneedleman: See discord

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -33,93 +34,131 @@ const useStyles = makeStyles({
...
+              editableSettings={editableSettings}
+              privacySetters={privacySetters}
+              setPrivacySettings={setPrivacySettings}
+              userId={userId}
+            />
+          )}
+        </Grid>
       </Grid>
-    </Grid>
+    </div>
```
mneedleman: Why is this wrapped in a div?

elp49: good eye, this was old code

api/models.py
rao53: ^

rao53 (commented):
Ignore ......

rao53 (dismissed):

frontend/src/components/profile/ProfileContent.tsx
```diff
@@ -0,0 +1,164 @@
...
+                      </Grid>
+                    </Grid>
+                  ))}
+                </Grid>
+              </Grid>
+            </Grid>
+          </Paper>
+        </Grid>
+      </Grid>
+    </div>
```
mneedleman: Why is this wrapped in a div?

frontend/src/components/profile/ProfileSidebar.tsx
```diff
@@ -0,0 +1,148 @@
...
+                      {title}
+                    </Typography>
+                  </Grid>
+                </Grid>
+              </Link>
+            </Grid>
+          </Grid>
+        ))}
+      </Grid>
+    </div>
```
mneedleman: Why is this wrapped in a div?

mneedleman (dismissed):
Not too excited about the div wrappers. I'd rather they be fragments.
Also your `Object.keys.map` can be replaced by a Ramda function, but is fine for now.

api/migrations/0010_auto_20210402_2245.py
```diff
@@ -0,0 +1,35 @@
...
+        data = json.load(fileData)
+        for interestData in data:
+            try:
+                category = interestData["category"]
+                title = interestData["title"]
+                interest = Interest(category=category, title=title)
+                interest.save()
+            except KeyError:
+                # Interest does not follow category/title structure.
+                pass
```
jcarrete5: Log this error so we know when it happens. This will tell us if there is a mistake in the interests file. Silently ignoring errors is almost never a good idea. 
```suggestion
            except KeyError as e:
                print(f"KeyError reading interests.json: {e}")
```

api/models.py
```diff
@@ -12,7 +12,41 @@ class Profile(models.Model):
...
+    gender_privacy = models.CharField(
+        max_length=2, choices=Setting.choices, default=Setting.PRIVATE
+    )
+    interests_privacy = models.CharField(
+        max_length=2, choices=Setting.choices, default=Setting.PUBLIC
+    )
+
+
+class Interest(models.Model):
+    profiles = models.ManyToManyField(Profile)
```
jcarrete5, line 47: When I mentioned many-to-many, I assumed it would be added to the profile model not the interests model, but I don't think it actually matters.

frontend/src/FullPageRoutes/Profile.tsx
```diff
@@ -33,92 +34,125 @@ const useStyles = makeStyles({
   },
 })
 
+function parseBirthday(birth_date: Nullable<Date>): Nullable<Date> {
+  let birthday = null
+  if (birth_date !== null) {
+    birthday = new Date(birth_date)
+    // Typescript is zero indexed but django is not.
```
jcarrete5:
```suggestion
    // We have Django configured to store dates using local time.
    // Date objects assume the date is UTC when parsing so we add 1
    // to correct for the EST/EDT - UTC timezone difference
```

jcarrete5 (changes requested):

sb3958:
It looks good.

api/views.py
```diff
@@ -82,21 +84,70 @@ def post(self, request):
...
+
+
+class InterestRetrieveView(generics.RetrieveAPIView):
+    serializer_class = InterestSerializer
+    queryset = Interest.objects.all()
+
+    def get(self, request):
+        allInterests = Interest.objects.all()
+        data = self.serializer_class(allInterests, many=True).data
+        return response.Response(data=data, status=status.HTTP_200_OK)
```
mneedleman: Since you declared the queryset, you might as well use it

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -0,0 +1,377 @@
...
+    // Reset editable content values.
+    setFirstName(readonlyContent.first_name)
+    setLastName(readonlyContent.last_name)
+    setBirthDate(readonlyContent.birth_date)
+    setGender(readonlyContent.gender)
+    setInterests(readonlyContent.interests)
+  }
+
+  return (
+    <div>
```
mneedleman: This can just be a fragment right?

frontend/src/components/profile/PrivacySettings.tsx
```diff
@@ -0,0 +1,272 @@
...
+}
+
+const useStyles = makeStyles({
+  pageTitle: {
+    fontSize: "2rem",
+    fontWeight: "bold",
+  },
+  fieldsContainer: {
+    paddingTop: 40,
+    paddingBottom: 10,
```
mneedleman: No unitless css please. Yes it defaults to px, but I'd prefer we be explicit. Also I'd prefer rem

frontend/src/components/profile/PrivacySettings.tsx
```diff
@@ -0,0 +1,272 @@
...
+    // Reset editable privacy settings.
+    setFirstNamePrivacy(readonlySettings.first_name_privacy)
+    setLastNamePrivacy(readonlySettings.last_name_privacy)
+    setBirthDatePrivacy(readonlySettings.birth_date_privacy)
+    setGenderPrivacy(readonlySettings.gender_privacy)
+    setInterestsPrivacy(readonlySettings.interests_privacy)
+  }
+
+  return (
+    <div>
```
mneedleman: This can just be a fragment right?

frontend/src/utils/profile.tsx
```diff
@@ -1,81 +1,210 @@
...
 export type ProfileContentSetters = {
   setFirstName: SetStateDispatch<string>
   setLastName: SetStateDispatch<string>
-  setBirthDate: SetStateDispatch<Date>
-  setGender: SetStateDispatch<string>
-  setVideo: SetStateDispatch<Nullable<string>>
-  setId: SetStateDispatch<number>
-  setInterests: SetStateDispatch<string>
+  setBirthDate: SetStateDispatch<Date | null>
+  setGender: SetStateDispatch<GenderKey | null>
```
mneedleman: Why not use `Nullable`

mneedleman (changes requested):
Some minor things I'd like to see changed

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -0,0 +1,377 @@
...
+            >
+              <Grid item>
+                <Avatar src={""} className={classes.profileAvatar} />
+              </Grid>
+              <Grid item>
+                {editableContent.video ? (
+                  <video
+                    controls
+                    src={editableContent.video}
+                    className={classes.introVideo}
```
mneedleman: Instead of the raw video component, can you add a `className` to the `components/video/Video` component and use that?

frontend/src/components/profile/PrivacySettings.tsx
elp49: lots of pages and components currently use the default px unit. whatever we choose it should be consistent throughout the app

frontend/src/components/profile/EditableProfileContent.tsx
elp49: sounds good, the RecordPage also uses a raw video should we change it there as well?

api/views.py
jcarrete5: You don't need this get method at all. You can just delete the get method and extend `generics.ListAPIView`. This is duplicated code

api/views.py
```diff
@@ -83,21 +85,70 @@ def post(self, request):
...
+
+class InterestRetrieveView(generics.RetrieveAPIView):
+    serializer_class = InterestSerializer
+    queryset = Interest.objects.all()
+
+    def get(self, request):
+        allInterests = self.get_queryset()
+        data = self.serializer_class(allInterests, many=True).data
+        return response.Response(data=data, status=status.HTTP_200_OK)
+
```
```suggestion
class InterestRetrieveView(generics.ListAPIView):
    serializer_class = InterestSerializer
    queryset = Interest.objects.all()
```

jcarrete5 (changes requested):

api/views.py
elp49: I already submitted these changes

jcarrete5: too fast for me.

jcarrete5 (dismissed):

rao53 (dismissed):

frontend/src/components/profile/PrivacySettings.tsx
mneedleman: We are choosing to have units on all css.

mneedleman (changes requested):

frontend/src/components/profile/PrivacySettings.tsx
elp49: do you want to add that to this pr or create another one for it

mneedleman: For the files in this PR, change it. We can fix others as they come up

mneedleman (approved):

PR #78, opened 2021-03-30
Updated app sidebar

elp49:
1Capture
2Capture
3Capture


elp49:
Profile sidebar will need to be removed and privacy settings moved to a full page after the privacy settings PR is merged. This would change the Profile and Privacy views

frontend/src/components/app/AppHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { useUser } from "../../utils/context"
+
+const useStyles = makeStyles({
+  header: {
+    backgroundColor: "#cd0b2d",
+    height: "3rem",
+    padding: "0 1rem",
+    position: "fixed",
+    left: 0,
+    top: 0,
```
mneedleman: No unitless css please

mneedleman (dismissed):

frontend/src/components/app/AppHeader.tsx
elp49: this defaults to px

sb3958 (dismissed):

AyalaCnaan (commented):
Possibly a stupid question, but should we consider moving the 'contact us' to this sidebar?

jcarrete5 (dismissed):

jcarrete5 (approved):

AyalaCnaan:
> Possibly a stupid question, but should we consider moving the 'contact us' to this sidebar?

Even if it is a stupid question, I would like an answer.

jcarrete5 (dismissed):

elp49:
> > Possibly a stupid question, but should we consider moving the 'contact us' to this sidebar?
> 
> Even if it is a stupid question, I would like an answer.

yeah the contact page is also in the sidebar

frontend/src/components/LoginForm.tsx
```diff
@@ -60,9 +62,29 @@ export default function LoginForm({
   const history = useHistory()
 
   const { from } = location.state || { from: { pathname: "/" } }
+  const [isOpen, setOpen] = React.useState<boolean>(false)
+  const [type] = useState<NotificationType["type"]>("error")
+  const [message] = useState("Email and/or password entered incorrectly")
```
jcarrete5: These states never change and can therefore just be constant.
```suggestion
  const type: NotificationType["type"] = "error"
  const message = "Email and/or password entered incorrectly"
```

frontend/src/components/SignupForm.tsx
```diff
@@ -24,9 +26,36 @@ export default function SignupForm({
   const location = useStatefulLocation()
   const { register } = useUser()
   const { from } = location.state || { from: { pathname: "/" } }
+  const [isModalOpen, setIsModalOpen] = React.useState<boolean>(false)
+  const [isNotifOpen, setNotifOpen] = React.useState<boolean>(false)
+  const [type] = React.useState<NotificationType["type"]>("error")
+  const [message] = React.useState("Something went wrong!")
```
jcarrete5: These states never change and can therefore just be constant.
```suggestion
  const type: NotificationType["type"] = "error"
  const message = "Something went wrong!"
```

jcarrete5 (changes requested):

frontend/src/components/LoginForm.tsx
elp49: this isn't a file I touched, I'm not sure I should make this change in this PR

jcarrete5: Weird. I don't know then. The behavior should be the same either way.

jcarrete5 (dismissed):

jcarrete5 (approved):

mneedleman (approved):

PR #80, opened 2021-04-09
Profile Picture Upload

elp49:


frontend/src/FullPageRoutes/Matching.tsx
```diff
@@ -168,7 +168,7 @@ export default function MatchingPage(): JSX.Element {
                   isOpen={introVisible}
                   senderId={match.id}
                   handleClose={() => setVisible(false)}
-                  intro
+                  intro={true}
```
mneedleman: This isn't necessary

frontend/src/components/AbstractModal.tsx
```diff
@@ -13,32 +14,42 @@ import * as React from "react"
...
+  },
+  close: {
+    position: "absolute",
+    top: 0,
+    right: 0,
+  },
+  title: {
+    margin: "25px 50px",
+    padding: 0,
   },
```
mneedleman: Was something wrong with AbstractModal? It seemed to work fine to me

api/views.py
```diff
@@ -190,27 +190,75 @@ def post(self, request, format=None):
...
         if serializer.is_valid():
+            # Save new intro video.
             serializer.save()
-            return response.Response(serializer.data, status=status.HTTP_201_CREATED)
-        return response.Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+            res = response.Response(serializer.data, status=status.HTTP_201_CREATED)
+        else:
+            res = response.Response(
+                serializer.errors, status=status.HTTP_400_BAD_REQUEST
+            )
```
mneedleman: What was wrong with the original version? I don't see a difference here functionally

frontend/src/components/AppFooter.tsx
```diff
@@ -10,6 +10,7 @@ const useStyles = makeStyles({
     position: "fixed",
     left: 0,
     bottom: 0,
+    zIndex: 1,
```
mneedleman: Why do these need a z index?

api/serializers.py
```diff
@@ -12,22 +12,46 @@
...
+def getVideoInstance(sender, receiver):
+    videoInstance = None
+    allVideoInstances = []
+
+    try:
+        # Get video instance.
+        allVideoInstances = Video.objects.filter(sender=sender, receiver=receiver)
+    except Exception as e:
+        # Video instance does not exist.
+        print(e)
```
mneedleman: This will never throw an Exception. <Model>.objects.filter returns an empty array if it doesn't find anything.

api/serializers.py
```diff
@@ -12,22 +12,46 @@
...
+
+    # Test if there exists an instance.
+    numVideos = len(allVideoInstances)
+    if numVideos == 1:
+        videoInstance = allVideoInstances[0]
+    # Test if there exists more than one instance.
+    elif numVideos > 1:
+        # TODO: should we double check how many intro videos there are here
+        # or not bother but delete old intro videos as they are updated?
+        videoInstance = allVideoInstances[numVideos - 1]
```
mneedleman: For an array of one element, `allVideoInstances[-1]` is equivalent to `allVideoInstances[0]`

api/serializers.py
```diff
@@ -111,11 +135,17 @@ def read_video(video):
     @param video: A FieldFile representing the video.
     Commonly found on Video.video_file
     """
+
+    videoData = None
     try:
-        return video.read().decode()
+        videoData = video.read().decode()
```
mneedleman: This should never fail. Were you encountering failures?

api/serializers.py
```diff
@@ -12,22 +12,46 @@
...
-        finally:
-            rep["video"] = video
-            return rep
+
+        videoData = None
+        videoInstance = getVideoInstance(instance.user, instance.user)
+        if videoInstance is not None:
+            videoData = read_video(videoInstance.video_file)
+            # TODO: if videoData comes back none, then video file is lost,
+            # should we delete the video instance?
```
mneedleman: I'm not sure this needed to be pulled out into a separate function, since it's still only used once.

api/urls.py
```diff
@@ -9,7 +9,7 @@
     path("deleteaccount/", views.UserAccountDeleteView.as_view()),
     path("videos/", views.VideoListCreate.as_view()),
     path("video/<int:sender>/", views.VideoRetrieveView.as_view()),
-    path("intro/<int:profile_id>/", views.IntroVideoRetrieveView.as_view()),
+    path("intro/<int:match>/", views.IntroVideoRetrieveUpdateView.as_view()),
```
mneedleman: Why the rename? I'd argue profile_id is better than match because it's more generic

api/views.py
```diff
@@ -190,27 +190,75 @@ def post(self, request, format=None):
...
+        return video
+
+    def get(self, request, match):
+        video = self.get_object(match)
+
+        # Test if an intro video exists.
+        res = None
+        if video is not None:
+            print("video is not none")
+            print(video)
```
mneedleman: Let's leave the print statements out

api/views.py
```diff
@@ -190,27 +190,75 @@ def post(self, request, format=None):
...
+            serializer = self.serializer_class(video)
+            res = response.Response(data=serializer.data, status=status.HTTP_200_OK)
+        else:
+            print("video is none")
+            res = response.Response(status=status.HTTP_404_NOT_FOUND)
```
mneedleman: If the ultimate goal is to 404, then why not use `get_or_404`?

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -37,14 +40,15 @@ function Preview({ stream }: PreviewProps) {
...
+  history: History,
+  route: VideoRoutes
 ) {
   if (blob) {
     const reader = new FileReader()
     reader.readAsDataURL(blob)
     reader.onload = () => {
       request({
-        path: "/api/videos/",
+        path: `/api/${route}/`,
```
mneedleman: I understand that the Intro video route was actually unused before, but I don't understand why we need it now.

api/views.py
```diff
@@ -190,27 +190,75 @@ def post(self, request, format=None):
...
+
+            res = response.Response(
+                data=serializer.data, status=status.HTTP_201_CREATED
+            )
+        else:
+            res = response.Response(
+                data=serializer.errors, status=status.HTTP_400_BAD_REQUEST
+            )
+
+        return res
```
mneedleman: The only difference between `IntroVideoRetrieveUpdateView.post` and `VideoListCreate.post` is the video deletion. Why not just check if `sender == receiver` and run the cleanup if true in the latter? No need to duplicate code that way

frontend/src/App.tsx
```diff
@@ -44,7 +44,7 @@ export default function App(): JSX.Element {
             <Route path="/tokens">
               <TokenPage />
             </Route>
-            <ProtectedRoute path="/record/:receiver">
+            <ProtectedRoute path="/record/:receiver/:type?">
```
mneedleman: Rather than adding extra url params (which I'm not really enthusiastic about doing), we can infer the type from the information provided by `useUser`. The returned user object has an id, and we can check if it's the same as the receiver id

frontend/src/components/AbstractModal.tsx
```diff
@@ -13,32 +14,42 @@ import * as React from "react"
...
+    margin: 0,
+  },
+  close: {
+    position: "absolute",
+    top: 0,
+    right: 0,
+  },
+  title: {
+    margin: "25px 50px",
+    padding: 0,
```
mneedleman: No unitless CSS

frontend/src/components/AbstractModal.tsx
```diff
@@ -13,32 +14,42 @@ import * as React from "react"
...
-    padding: theme.spacing(2),
+    top: 0,
+    left: "50%",
+    transform: "translate(-50%, 25%)",
+    margin: 0,
+  },
+  close: {
+    position: "absolute",
+    top: 0,
+    right: 0,
```
mneedleman: Why does the close button need absolute positioning?

frontend/src/components/AbstractModal.tsx
```diff
@@ -13,32 +14,42 @@ import * as React from "react"
...
 type ModalType = {
   title: string
-  description: string
+  description?: string
   confirmText?: string
   cancelText?: string
   isModalOpen: boolean
   handleConfirm: () => void
   handleCancel: () => void
+  children?: JSX.Element
```
mneedleman: I think the standard is to have `children` have the type `React.ReactNode`

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -253,38 +348,86 @@ export default function EditableProfileContent({
...
+                <Grid
+                  container
+                  alignItems="center"
+                  justify="center"
+                  className={`${classes.iconContainer} ${classes.cameraContainer}`}
+                >
+                  <Grid item className={classes.cameraIconItem}>
+                    <CameraAltIcon className={classes.cameraIcon} />
+                  </Grid>
+                </Grid>
```
mneedleman: This is starting to feel like it needs some organizational refactoring.

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -253,38 +348,86 @@ export default function EditableProfileContent({
...
+                  to={`/record/${userId}`}
+                >
                   <Video
                     src={editableContent.video}
-                    className={classes.introVideo}
+                    className={classes.video}
                   />
-                ) : (
+                  <div className={classes.blur}></div>
+                  <div className={classes.overlay}></div>
```
mneedleman: What are the empty divs for?

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -253,38 +348,86 @@ export default function EditableProfileContent({
...
+              >
+                <Avatar
+                  src={
+                    editableContent.image_file !== null
+                      ? editableContent.image_file
+                      : ""
+                  }
+                  className={classes.picture}
+                />
+                <div className={`${classes.overlay} ${classes.circle}`}></div>
```
mneedleman: Why the empty div?

frontend/src/components/profile/UploadImageModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    margin: 5,
+  },
+  filename: {
+    marginTop: 10,
+  },
+  error: {
+    minHeight: "1.5rem",
+    marginTop: 5,
+    color: "red",
+  },
```
mneedleman: No unitless CSS

frontend/src/components/profile/UploadImageModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const [inputImageFile, setInputImageFile] = React.useState<Nullable<File>>(
+    null
+  )
+
+  const handleChangeImageInput = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const fileList = e.currentTarget.files
+    const file = fileList && fileList.length > 0 ? fileList[0] : inputImageFile
+    setInputImageFile(file)
+    setErrorMessage("")
+  }
```
mneedleman, line 46: Is there a way to limit to just 1 file?

frontend/src/components/profile/UploadImageModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+        method: "put",
+        body: formData,
+        isFormData: true,
+      })
+        .then((res) => {
+          updateProfilePicture(res.parsedBody.image_file)
+          closeModal()
+        })
+        .catch((error) => {
+          console.log(error)
```
mneedleman: Seems like a good time to set the error message

frontend/src/components/profile/UploadImageModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+  return (
+    <AbstractModal
+      title="Upload Your Profile Picture"
+      confirmText="Save"
+      cancelText="Cancel"
+      isModalOpen={isModalOpen}
+      handleConfirm={handleConfirmUpload}
+      handleCancel={closeModal}
+    >
+      <div>
```
mneedleman: No divs please

frontend/src/components/profile/UploadImageModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+            <label htmlFor="uploadImageButton">
+              <Button
+                variant="contained"
+                component="span"
+                className={classes.button}
+              >
+                Select File
+              </Button>
+            </label>
+          </Grid>
```
mneedleman: This grid should probably be marked with `item` and `container`. And the parent should be a `container`

frontend/src/components/profile/UploadImageModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    if (inputImageFile !== null) {
+      // Send file in form data.
+      const formData = new FormData()
+      formData.append("image_file", inputImageFile)
+      request<ProfileContentType>({
+        path: "/api/profile/",
+        method: "put",
+        body: formData,
+        isFormData: true,
+      })
```
mneedleman: Why not just send the image?

frontend/src/utils/fetch.tsx
```diff
@@ -50,17 +52,18 @@ export async function request<T>({
   body,
   headers = defaultRequestArgs.headers,
   parser = defaultRequestArgs.parser,
+  isFormData = false,
```
mneedleman: I would just add this to `defaultRequestArgs`

frontend/src/utils/fetch.tsx
```diff
@@ -50,17 +52,18 @@ export async function request<T>({
...
 }: requestArgs<T>): Promise<HttpResponse<T>> {
   const csrfToken = Cookies.get("csrftoken")
   const args: RequestInit = {
     method: method,
     headers: {
-      ...headers,
+      ...(!isFormData && headers),
```
mneedleman: Why was this necessary?

frontend/src/utils/fetch.tsx
```diff
@@ -50,17 +52,18 @@ export async function request<T>({
...
       //conditionally adds the X-CSRFToken property
       ...(csrfToken && { "X-CSRFToken": csrfToken }),
     },
     //conditionally adds the body
-    ...(body && { body: JSON.stringify(body) }),
+    ...(body && isFormData ? { body: body } : { body: JSON.stringify(body) }),
```
mneedleman: I wonder if this deserves the same treatment as the response parser

mneedleman (changes requested):
There's a lot to review here. In addition to my specific feedback, could you provide more comments on the frontend and more screenshots? Thanks

api/serializers.py
jcarrete5: Since we are treating sender=user and receiver=user as a special case for storing the intro video, a constraint on the Video model prohibiting multiple videos with the same sender and receiver might be useful to keep the integrity of the database. This could be difficult to do, however.

If the goal of this code is to get the last video, then you can just do `allVideoInstances[-1]` like Mike mentioned. But can you be sure that the last video is the right (most recent) video to use? Maybe order by the date in the query above?

api/serializers.py
jcarrete5: Yea, the try-finally only exists to close the file after it is read; It doesn't catch any errors.

api/views.py
```diff
@@ -190,27 +190,75 @@ def post(self, request, format=None):
...
 
-    def get_object(self, profile_id):
-        return Video.objects.get(Q(sender=profile_id) & Q(receiver=profile_id))
+    def get_object(self, userId):
+        video = None
+        try:
+            video = Video.objects.get(sender=userId, receiver=userId)
+        except Exception:
+            # No pre-existing video.
+            pass
```
jcarrete5: The assertion of "No pre-existing video" in the except clause is not accurate. `Video.objects.get` can also throw an error if there are multiple results for the query. Do you intend to catch that error as well? This is not clear given the comment.

What is supposed to happen when there are multiple intro videos? Can there multiple intro videos? Should `None` be returned for that case?

api/views.py
jcarrete5: If something should be printed, use [loggers](https://docs.djangoproject.com/en/3.1/topics/logging/).

frontend/src/components/profile/EditableProfileContent.tsx
jcarrete5: Looks like the CSS classes on the divs explain why. They don't have any content other than displaying stuff with CSS.

frontend/src/components/profile/UploadImageModal.tsx
jcarrete5: HTML input elements have a `files` property that is a `FileList`. Afaik, this property should only [contain at most one file](https://html.spec.whatwg.org/multipage/input.html#file-upload-state-(type=file)) unless the `multiple` attribute is specified on the input element.

frontend/src/components/profile/UploadImageModal.tsx
jcarrete5: Yea, FormData feels a bit weird to use here considering all the other options. The path of least resistance would probably be to simply encode the image as a data URI like we do for the videos. Ideally, we'd just send the raw binary blob as the body.

jcarrete5 (changes requested):

api/serializers.py
```diff
@@ -12,22 +12,46 @@
 class ProfileSerializer(serializers.ModelSerializer):
```
rao53, line 12: Shouldn't this also be returning the image instance so it can be displayed? I know you mentioned that Django doesn't like loading static files... Not sure if this makes sense but could we treat the image like a video?

AyalaCnaan (commented):
I'm very confused by what's going on here. I think I need a step by step explanation of the code changes before I can say anything useful.

api/serializers.py
elp49: yeah we might not have a choice if we can't get media files to be treated like static files without using django templates. I can make this change if everyone else agrees.

api/serializers.py
elp49: this is a "clean clode" practice, it makes it so the code does exactly as it reads and declutters the rep function

api/serializers.py
elp49: yeah I don't think I did a great job explaining myself in my comments in the elif clause below. I think a `Video.objects.get` is what we really want here since there should probably only be one intro video at a time for any user. I guess it doesn't matter either way but a get makes more sense to me. the question is how do we want to enforce the single intro video restraint

api/serializers.py
elp49: I don't believe django supports negative indexing. we could use `.order_by('-id')[0:1]` as a workaround but I thought this approach looked nicer, although ording by the id would probably mean we don't need to assume the last video is the most recent.

api/serializers.py
elp49: I tested what would happen if video file got deleted from the media/video dir somehow and the endpoint fails but this is a workaround

api/urls.py
elp49: that's why I changed it to match to be more specific. The only time a profile id is passed to this endpoint is from the matching page to retrieve a matches intro video

api/views.py
elp49: yeah no difference, when I learned to write endpoints I read that multiple return statements can be misleading for someone who is trying to figure out what is being returned. but this view is small anyway so it doesn't make much of a difference

api/views.py
elp49: that's a good question, I wanted to hear what you all think. if we old intros after a new one is posted then there should only ever be 1 or 0. my only fear is if somewhere down the line a bug pops up that allows multiple intros and if that ever happens then should the intro get function check for multiple videos and delete old ones?

api/views.py
elp49: no problem, I forgot to take this out after testing

api/serializers.py
jcarrete5: Looking through the source for `QuerySet.__getitem__`, it appears that negative indexing is explicitly not allowed on QuerySets. You could convert the queryset to a list like `list(allVideoInstances)[-1]`

elp49: ok that could work too, but we still might need to worry about the ordering like you said. 

Do you guys think the get intro video function AND the post intro function should delete any possible old ones or just the post?

api/serializers.py
elp49: actually I just noticed I never committed my change to make this function `getIntroVideo(userId)` where it only passes the user id

api/views.py
elp49: sounds good

api/views.py
elp49: yeah that's how I had it originally but then moved it because I thought it could be misleading if someone was looking at the code and saw intro video post and get functions came from two different views. I don't care either way, what do you think?

frontend/src/App.tsx
elp49: ok, I thought you wanted to keep the intro video view. If we remove the type then we could combine the intro video view with video list and then the ViewVideo component could just hit api/video. what do you think?

frontend/src/FullPageRoutes/RecordPage.tsx
elp49: it was used by the ViewVideo component as you pointed out. but I agree, I don't think we need it

frontend/src/components/AbstractModal.tsx
elp49: ok

frontend/src/components/AbstractModal.tsx
elp49: it doesn't need it and there's simple ways around it, but absolute makes sense to me in this context and is the way most pop up windows do it from what I've seen

frontend/src/components/AbstractModal.tsx
elp49: there was empty space at the top and the content was crowded on the bottom. especially when I added the child node things looked off

frontend/src/components/AppFooter.tsx
elp49: the header and footer have position fixed which means that any element with position relative will be shown on top of it. you can see this with the avatar components on profile page

frontend/src/components/profile/UploadImageModal.tsx
elp49: yeah what jason said. checking if the length is greater than zero here will only overwrite the current `inputImageFile` if a new one has been selected

frontend/src/components/profile/UploadImageModal.tsx
elp49: sure I can look into this, most people uploading images using django also use form data

frontend/src/components/profile/UploadImageModal.tsx
elp49: I thought I mentioned this already that the current fetch wrapper we are using does not return an error message. it returns a generic exception

frontend/src/components/profile/UploadImageModal.tsx
elp49: ok

frontend/src/components/profile/UploadImageModal.tsx
elp49: I can add that in if you'd really like but in my opinion all of that makes the code more cluttered but produced the same look in the browser. It would mean we add `container` to the parent and `container justify="center" item` to each child but in the end it all looks the same

frontend/src/utils/fetch.tsx
elp49: if we decide not to use form data then I can just take this out

frontend/src/utils/fetch.tsx
elp49: the correct form data request headers will be added by default

api/views.py
mneedleman: I'm not convinced this is inside the scope of your PR

frontend/src/components/AppFooter.tsx
mneedleman: This is outside the scope of your PR

api/serializers.py
mneedleman: I'm not convinced this is inside the scope of your PR

api/serializers.py
mneedleman: Isn't there a `myQuerySet.last()`?

mneedleman: I'm also not convinced that video deletion is within the scope of this PR.

api/serializers.py
mneedleman: I'm not convinced that video deletion is inside the scope of this PR

api/serializers.py
mneedleman: When are you planning on doing that?

api/urls.py
mneedleman: Is your work using this endpoint then? Or is it still just the matching page?

api/views.py
mneedleman: I thin it's cleaner to keep it in one. I'd rather stay DRY

frontend/src/App.tsx
mneedleman: I'm not sure what this has to do with the view. I'm saying that you don't need the extra url parameter here

frontend/src/components/AbstractModal.tsx
mneedleman: Editing the positioning of the close button in the modal is outside the scope of this PR

frontend/src/components/profile/EditableProfileContent.tsx
mneedleman: I'd rather handle this through JSX and TS than CSS.

frontend/src/components/profile/UploadImageModal.tsx
mneedleman: That's not at all what I'm saying. You declare a state `errorMessage` using `React.useState`. Use that

frontend/src/utils/fetch.tsx
mneedleman: I'm not following you

api/views.py
mneedleman: Github comments are not a good place for design discussion

api/serializers.py
mneedleman: Let's have this discussion in discord or in a meeting

mneedleman:
I tried checking out the PR and uploading an image, but was unsuccessful. It seems like the upload was ok, but it did not retrieve correctly. Can you look into this further?

api/serializers.py
jcarrete5: So `QuerySet.last` is probably the best option. The QuerySet would still need to be sorted or we need to be sure there can only ever be one making sorted unnecessary.

I also agree that this is probably out of scope. Why do you need to make changes to how the intro video is handled? I think recording this as a bug in clickup suffices.

frontend/src/components/profile/UploadImageModal.tsx
elp49: sure I can add that, but the only message we can set is "Failed to update profile" because the fetch wrapper does not return any details

mneedleman: That's fine

api/serializers.py
elp49: see the latest updates below

api/serializers.py
```diff
@@ -13,45 +13,30 @@ class ProfileSerializer(serializers.ModelSerializer):
...
-        if videoInstance is not None:
-            videoData = read_video(videoInstance.video_file)
-            # TODO: if videoData comes back none, then video file is lost,
-            # should we delete the video instance?
-
-        rep["video"] = videoData
-        return rep
+        # Replace image file path with the file data.
+        image = instance.image
+        if image != "":
```
mneedleman, line 18: This can be null right?

jcarrete5:
> I tried checking out the PR and uploading an image, but was unsuccessful. It seems like the upload was ok, but it did not retrieve correctly. Can you look into this further?

I thought I was having the same issue but I just didn't save after choosing an image. Did you press the save button after changing the image? It's a bit confusing because you click save when choosing the image and press save again to commit the changes. It was working fine after I did that.

api/serializers.py
```diff
@@ -12,6 +12,15 @@
 class ProfileSerializer(serializers.ModelSerializer):
     def to_representation(self, instance):
         rep = super().to_representation(instance)
+
+        # Replace image file path with the file data.
+        image = instance.image
+        if image != "":
```
jcarrete5, line 18:
```suggestion
        if hasattr(image, "file"):
```
The existing implementation returns an internal server error if the instance does not have an associated file for the image.

api/views.py
```diff
@@ -115,15 +117,37 @@ def get(self, request):
...
+
+        if imageData is not None:
+            # Data is prepended by "data:image/*;base64,".
+            # Encode converts from string to bytes.
+            data_bytes = imageData.encode()
+
+            # Create the image content file.
+            created_at = now()
+            fname = "%d_%s" % (request.user.id, created_at)
```
```suggestion
            fname = f"{request.user.id}_{created_at}"
```
Percent format string is fine but I suggest using f-strings. They are newer and arguably easier to read.

jcarrete5 (changes requested):

jcarrete5 (dismissed):

rao53 (dismissed):

frontend/src/components/profile/EditableProfileContent.tsx
elp49: I'm not familiar with how to do that for this specific instance, can you suggest anything?

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -52,17 +55,62 @@ const useStyles = makeStyles({
...
-  image: {
-    width: "inherit",
-    height: "inherit",
-    objectFit: "cover",
+  overlay: {
+    height: "100%",
+    width: "100%",
+    position: "absolute",
+    top: 0,
+    right: 0,
```
mneedleman: 1. No unitless css please. 2. I tried this without these attributes, and it worked just fine.

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -40,6 +42,7 @@ type EditableProfileContentProps = {
 const MAX_FIRST_NAME_LENGTH = 50
 const MAX_LAST_NAME_LENGTH = 50
 
+const pictureSize = 200
```
mneedleman, line 46: You define a picture size, but its use is inconsistent

api/serializers.py
```diff
@@ -12,6 +12,15 @@
 class ProfileSerializer(serializers.ModelSerializer):
     def to_representation(self, instance):
         rep = super().to_representation(instance)
+
+        # Replace image file path with the file data.
+        image = instance.image
+        if hasattr(image, "file"):
```
mneedleman: This raises a ValueError if there is no file

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -52,17 +55,62 @@ const useStyles = makeStyles({
...
+    opacity: "0%",
+    "&:hover": {
+      opacity: "25%",
+    },
   },
-  uploadImage: {
+  iconContainer: {
+    position: "absolute",
+    bottom: 3,
+    right: 3,
```
mneedleman: No unitless css.

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -52,17 +55,62 @@ const useStyles = makeStyles({
...
+    },
+    cursor: "pointer",
+    boxShadow: "0 4px 8px 0 rgb(0 0 0 / 20%), 0 6px 20px 0 rgb(0 0 0 / 19%)",
+  },
+  blur: {
+    height: "25%",
+    width: "100%",
+    position: "absolute",
+    bottom: 0,
+    left: 0,
```
mneedleman: No unitless css.

frontend/src/components/profile/EditableProfileContent.tsx
mneedleman: This can be replaced with a 75% opacity on hover on the picture css class.

mneedleman (changes requested):
This still needs some work. I still have issues uploading a picture. I get a 500 on the profile endpoint

api/serializers.py
jcarrete5: I'm having trouble reproducing this. Could you summarize the steps you took to cause this to throw a value error?

elp49: It throws `ValueError: The 'image' attribute has no file associated with it.` when no image has been saved to it yet

elp49: is there anything wrong with using `image != ""` instead? It works as expected

jcarrete5: That was not working for me which is why I suggested the change

frontend/src/components/profile/EditableProfileContent.tsx
elp49: that would make it so you can see through the picture, do you think that looks better?

mneedleman (dismissed):
I still feel like you're wrestling with the css unnecessarily, but I don't have time to explore it.

jcarrete5 (dismissed):

rao53 (dismissed):

jcarrete5 (approved):

mneedleman (approved):

PR #81, opened 2021-04-11
Matching V2

jcarrete5:
image

### DIFF
--- a/frontend/src/FullPageRoutes/Matching.tsx
+++ b/frontend/src/FullPageRoutes/Matching.tsx
@@ -33,7 +33,7 @@ type Match = {
   birth_date: string
   email: string
   phone_number: string
-  interests: string
+  interests: Interest[]
   has_intro: boolean
 }
 
@@ -199,6 +199,10 @@ function useQueue<T>(): [T, () => void, EnqueueFunc<T>, () => void] {
   return [items[0], nextItem, enqueueItems, clear]
 }
 
+function interestString(interest: Interest) {
+  return `${interest.category}::${interest.title}`
+}
+
 export default function MatchingPage(): JSX.Element {
   const [match, next, enqueue, clearQ] = useQueue<Match>()
   const [filters, dispatch] = React.useReducer(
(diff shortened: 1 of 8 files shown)


AyalaCnaan (commented):
As far as I can tell this is good. I am confused by the screen shot, why is email on there?

jcarrete5:
> As far as I can tell this is good. I am confused by the screen shot, why is email on there?

The UI is very much subject to change. As of right now, the match card doesn't respect user privacy so the email shows up. I have noted this as a bug in [clickup](https://app.clickup.com/t/t95jvw).

frontend/src/FullPageRoutes/Matching.tsx
```diff
@@ -6,11 +6,59 @@ import {
...
+  last_name: string
+  gender: string
+  birth_date: string
+  email: string
+  phone_number: string
+  interests: Interest[]
+  has_intro: boolean
+}
+
+type InterestID = number
```
mneedleman, line 40: Why is this necessary?

frontend/src/FullPageRoutes/Matching.tsx
```diff
@@ -6,11 +6,59 @@ import {
...
+  OTHER = "O",
+}
+
+class FilterState {
+  static readonly ageLimits = { min: 18, max: 130 }
+  readonly minAge = FilterState.ageLimits.min
+  readonly maxAge = FilterState.ageLimits.max
+  readonly genders = Object.values(Gender)
+  readonly interests: Interest[] = []
+}
```
mneedleman, line 60: I'm pretty hesitant to take an OOP approach, but I can't deny that's concise and clean. However, as a 131 year old, I feel discriminated against

frontend/src/utils/fetch.tsx
```diff
@@ -65,3 +65,36 @@ export async function request<T>({
...
+        } else {
+          throw new Error(`Failed to encode key/value pair: ${key}=${val}`)
+        }
+        break
+      default:
+        throw new Error(`Failed to encode key/value pair: ${key}=${val}`)
+    }
+  }
+  return "?" + params.join("&")
+}
```
mneedleman, line 100: If we had more time, I'd say we should explore how RamdaJS might help clean this up. But looks good

mneedleman (approved):

frontend/src/FullPageRoutes/Matching.tsx
jcarrete5: It's not. I was trying to avoid using the `Interest` type in the filter state so it would be easier to encode in the URL. I don't need it anymore but I didn't feel it was worth removing

mneedleman: Fair

AyalaCnaan (approved):

PR #82, opened 2021-04-11
Forgotpassword

rao53:


rao53:
High-level overview:
- input your valid email address
- it sends you a new link only accessible only by user to reset your password
- Password is reset

Side note: We need to change the email that is sending the account verification email. I also need to share the API-KEY with the team.

frontend/src/App.tsx
```diff
@@ -44,6 +46,13 @@ export default function App(): JSX.Element {
             <Route path="/tokens">
               <TokenPage />
             </Route>
+            <Route path="/forgotpassword">
+              <ForgotPassword />
+            </Route>
+            <Route path="/verification/:uidb64/:token">
+              {/* <Route path="/verification?uidb64=:uidb64&token=:token"> */}
```
mneedleman: Might as well remove the commented line

api/views.py
```diff
@@ -60,6 +68,80 @@ def delete(self, request):
...
+            )
+            reset_url = "http://" + domain + link
+            try:
+                send_mail(
+                    "Password reset request",
+                    "Hi "
+                    + currentUser.username
+                    + " Please use this link to reset your account\n"
+                    + reset_url,
+                    "olawinridwan18@gmail.com",  # Need to change to a conteo email
```
mneedleman: Is this parameter the email it is sent from?

mneedleman (changes requested):

api/views.py
jcarrete5: Yes.

frontend/src/App.tsx
rao53: gotcha

frontend/src/FullPageRoutes/ResetPassword.tsx
```diff
@@ -0,0 +1,170 @@
...
+      parser: parseIdentity,
+    })
+      .then(() => {
+        setType("success")
+        setIsOpen(true)
+        setMessage("Password changed successfully")
+        setTimeout(function () {
+          console.log("I changed")
+          history.push("/")
+        }, 2000)
```
mneedleman, line 99: Why is the timeout necessary?

api/views.py
mneedleman: Might as well use our 7th member

api/views.py
```diff
@@ -60,6 +68,80 @@ def delete(self, request):
...
+                    [username],
+                    fail_silently=False,
+                )
+                # currentUser.set_password("rao53") #Use newPassword her
+                # currentUser.save()
```
mneedleman: Don't need these

mneedleman (changes requested):

api/views.py
AyalaCnaan: Just for my learning, why don't we need these?

frontend/src/FullPageRoutes/ForgotPassword.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { NotificationType } from "../components/Notification"
+import { parseIdentity, request } from "../utils/fetch"
+
+export type UserAccounts = {
+  first_name: string
+  last_name: string
+  username: string
+  password: string
+  email: string
+  successor: string
```
AyalaCnaan: For the sake of my getting a better understanding of what's going on, why are we pulling out first name, last name and successor?

AyalaCnaan (dismissed):

frontend/src/FullPageRoutes/ResetPassword.tsx
rao53: Just to change the page to the Home Screen after changing the password. It might be unnecessary to be honest.

api/views.py
rao53: It's commented out. It was just something I was playing around with when I was doing the implementation.

mneedleman (approved):

api/views.py
```diff
@@ -60,6 +68,78 @@ def delete(self, request):
...
+                    "Hi "
+                    + currentUser.username
+                    + " Please use this link to reset your account\n"
+                    + reset_url,
+                    "conteobot@gmail.com",
+                    [username],
+                    fail_silently=False,
+                )
+            except OSError:
+                return response.Response(status=status.HTTP_400_BAD_REQUEST)
```
elp49, line 108: does this require a certain API_KEY? I got a `OSError:  Connection unexpectedly closed` when testing

elp49 (commented):
does the send_

api/views.py
rao53: I just texted it to you

mneedleman: Can you post it in the discord

elp49: with the correct api key this worked as expected for me

elp49 (approved):

jcarrete5 (approved):

PR #86, opened 2021-04-19
Form Function

sb3958:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "c:\\Users\\Sharelle\\Desktop\\Senior Design\\conteo\\venv\\Scripts\\python.exe"
```
mneedleman: `.vscode` should probably be in the `.gitignore`. We don't need this in the repo

mneedleman (changes requested):
We should discuss whether these pages should be authenticated routes or not.

api/urls.py
```diff
@@ -21,4 +21,6 @@
     path("reports/", views.Reports.as_view()),
     path("accounts/", views.Accounts.as_view()),
     path("user/", views.UserAuthView.as_view()),
+    path("contact/", views.ContactUsView.as_view()),
+    path("feedback/", views.FeedbackView.as_view()),
```
elp49: these views do not exist

api/serializers.py
```diff
@@ -205,3 +214,15 @@ def to_representation(self, instance):
...
+class ContactUsSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = ContactUs
+        fields = ("email", "name", "message")
+
+
+class FeedbackSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Feedback
+        fields = ("reason", "message", "email")
```
elp49: I think these fields can be `fields = '__all__'`

frontend/src/FullPageRoutes/AccountsPage.tsx
```diff
@@ -69,7 +71,8 @@ export default function AccountsPage(): JSX.Element {
     editModeOn(!editMode)
   }
 
-  const handleSave = () => {
+  const handleSave = (e: any) => {
+    e.preventDefault().then(history.push("/"))
```
elp49: preventDefault doesn't return a value. do we really want to redirect after a user saves?

frontend/src/FullPageRoutes/ContactUs.tsx
```diff
@@ -50,13 +52,23 @@ const contactValue = {
...
     const { name, value } = e.target
     setValues({
       ...values,
       [name]: value,
     })
+    return request({
+      path: "/api/contact/",
+      method: "post",
+      body: contactValue,
+    })
```
elp49, line 68: the request should be in `handleSubmit`

frontend/src/FullPageRoutes/ContactUs.tsx
```diff
@@ -50,13 +52,23 @@ const contactValue = {
 export default function ContactUs(): JSX.Element {
   const [values, setValues] = useState(contactValue)
   const classes = useStyles()
+  const history = useHistory()
+
+  const handleSubmit = (e: any) => {
+    e.preventDefault().then(history.push("/help"))
```
elp49: preventDefault doesn't return a value.

frontend/src/FullPageRoutes/FeedbackPage.tsx
```diff
@@ -39,13 +41,23 @@ const feedBackValue = {
 export default function FeedbackPage(): JSX.Element {
   const [values, setValues] = useState(feedBackValue)
   const classes = useStyles()
+  const history = useHistory()
+
+  const handleSubmit = (e: any) => {
+    e.preventDefault().then(history.push("/help"))
```
elp49: preventDefault doesn't return a value.

frontend/src/FullPageRoutes/FeedbackPage.tsx
```diff
@@ -39,13 +41,23 @@ const feedBackValue = {
...
     const { name, value } = e.target
     setValues({
       ...values,
       [name]: value,
     })
+    return request({
+      path: "/api/feedback/",
+      method: "post",
+      body: feedBackValue,
+    })
```
elp49: the request should be in `handleSubmit`

elp49 (changes requested):

.vscode/settings.json
mneedleman: This still needs to be removed from the pushed files

api/models.py
```diff
@@ -140,3 +140,17 @@ class ReportType(models.TextChoices):
...
+    name = models.CharField(max_length=50, blank=True)
+    message = models.CharField(max_length=2000, null=True)
+
+
+class Feedback(models.Model):
+    FEEDBACK_REASON = (("W", "Website"), ("Pp", "Penpal"), ("P", "Performance"))
+
+    reason = models.CharField(max_length=2, choices=FEEDBACK_REASON, null=True)
+    message = models.CharField(max_length=2000, blank=True)
+    email = models.CharField(max_length=50, blank=True)
```
mneedleman: I think there needs to be a generated migration for this to work properly

api/views.py
```diff
@@ -416,3 +418,30 @@ class Accounts(generics.RetrieveAPIView):
...
+    serializer_class = ContactUsSerializer
+
+    def post(self, request):
+        email = request.data["email"]
+        name = request.data["name"]
+        message = request.data["message"]
+        serializer = self.serializer_class(
+            context={"email": email, "name": name, "message": message}
+        )
+        return response.Response(serializer.data, status=status.HTTP_201_CREATED)
```
mneedleman: I don't think you need to pull those out of the request and pass them in with the context. You should be able to pass the request

api/views.py
```diff
@@ -416,3 +418,30 @@ class Accounts(generics.RetrieveAPIView):
...
+
+    def post(self, request):
+        reason = response.data["reason"]
+        message = response.data["message"]
+        email = response.data["email"]
+        if User.objects.get(request.user):
+            serializer = self.serializer_class(
+                context={"reason": reason, "message": message, "email": email}
+            )
+            return response.Response(serializer.data, status=status.HTTP_201_CREATED)
```
mneedleman: I'm not sure the check for user is necessary

api/views.py
```diff
@@ -416,3 +418,30 @@ class Accounts(generics.RetrieveAPIView):
     def get(self, request):
         serializer = AccountSerializer(request.user)
         return response.Response(serializer.data, status=status.HTTP_200_OK)
+
+
+class ContactUs(generics.CreateAPIView):
+    serializer_class = ContactUsSerializer
```
mneedleman, line 453: Since the ContactUs page does not require the user to be logged in, we should add the AllowAny permission
```suggestion
class ContactUs(generics.CreateAPIView):
    serializer_class = ContactUsSerializer
    permission_clsses = (permissions.AllowAny)
```

frontend/src/FullPageRoutes/AccountsPage.tsx
mneedleman: I agree with Eddie

frontend/src/FullPageRoutes/ContactUs.tsx
```diff
@@ -51,12 +52,17 @@ export default function ContactUs(): JSX.Element {
...
     const { name, value } = e.target
     setValues({
       ...values,
       [name]: value,
     })
+    return request({
+      path: "/api/contact/",
+      method: "post",
+      body: contactValue,
+    }).then(e.preventDefault())
```
mneedleman: Typically `e.preventDefault()` is used to ensure that the click event does not propagate. What are you trying to do here?

mneedleman (changes requested):

frontend/src/FullPageRoutes/ContactUs.tsx
```diff
@@ -51,12 +52,17 @@ export default function ContactUs(): JSX.Element {
   const [values, setValues] = useState(contactValue)
   const classes = useStyles()
 
-  const setContactValue = (e: any) => {
```
elp49: why did you get rid of the set function? We will need both setContactValue and handleSubmit

frontend/src/FullPageRoutes/ContactUs.tsx
```diff
@@ -99,14 +105,14 @@ export default function ContactUs(): JSX.Element {
                   label="Your name"
                   name="name"
                   value={values.name}
-                  onChange={setContactValue}
+                  onChange={handleSubmit}
```
elp49: why did you change this to handleSubmit?

frontend/src/FullPageRoutes/FeedbackPage.tsx
```diff
@@ -74,7 +80,7 @@ export default function FeedbackPage(): JSX.Element {
                   id="reasons"
                   name="reasons"
                   value={values.reason}
-                  onChange={setFeedbackValue}
+                  onChange={handleSubmit}
```
elp49: this should be setFeedbackValue

frontend/src/FullPageRoutes/FeedbackPage.tsx
```diff
@@ -91,14 +97,14 @@ export default function FeedbackPage(): JSX.Element {
                   multiline
                   rowsMax={6}
                   value={values.message}
-                  onChange={setFeedbackValue}
+                  onChange={handleSubmit}
```
elp49: this should be setFeedbackValue

frontend/src/FullPageRoutes/FeedbackPage.tsx
```diff
@@ -91,14 +97,14 @@ export default function FeedbackPage(): JSX.Element {
...
                 />
                 <TextField
                   variant="outlined"
                   label="Your email"
                   name="email"
                   value={values.email}
-                  onChange={setFeedbackValue}
+                  onChange={handleSubmit}
```
elp49: this should be setFeedbackValue

frontend/src/FullPageRoutes/FeedbackPage.tsx
```diff
@@ -40,12 +41,17 @@ export default function FeedbackPage(): JSX.Element {
   const [values, setValues] = useState(feedBackValue)
   const classes = useStyles()
 
-  const setFeedbackValue = (e: any) => {
+  const handleSubmit = (e: any) => {
```
elp49: we will also need setFeedbackValue

frontend/src/FullPageRoutes/ContactUs.tsx
```diff
@@ -99,14 +105,14 @@ export default function ContactUs(): JSX.Element {
...
                 />
                 <TextField
                   variant="outlined"
                   label="Your email"
                   name="email"
                   value={values.email}
-                  onChange={setContactValue}
+                  onChange={handleSubmit}
```
elp49: this should be setContactValue

frontend/src/FullPageRoutes/ContactUs.tsx
```diff
@@ -115,7 +121,7 @@ export default function ContactUs(): JSX.Element {
                   multiline
                   rowsMax={6}
                   value={values.message}
-                  onChange={setContactValue}
+                  onChange={handleSubmit}
```
elp49: this should be setContactValue

frontend/src/FullPageRoutes/ContactUs.tsx
```diff
@@ -125,7 +131,7 @@ export default function ContactUs(): JSX.Element {
                   className={classes.submitButton}
                   size="large"
                   type="submit"
-                  href="#"
+                  onSubmit={handleSubmit}
```
elp49, line 137: good

elp49 (changes requested):

frontend/src/FullPageRoutes/ContactUs.tsx
sb3958: So just put the return request in the handleSubmit.

elp49: right, you don't want to make the request until the user selects the submit button, but setContactValue should be called in the onChange of each input

api/migrations/0011_contactus_feedback.py
```diff
@@ -0,0 +1,58 @@
...
+                        choices=[
+                            ("W", "Website"),
+                            ("Pp", "Penpal"),
+                            ("P", "Performance"),
+                        ],
+                        max_length=2,
+                        null=True,
+                    ),
+                ),
+                ("message", models.CharField(blank=True, max_length=2000)),
```
AyalaCnaan, line 54: Just curious why this limit on message length.

frontend/src/FullPageRoutes/AccountsPage.tsx
AyalaCnaan: Just curious, but what does it do?

api/migrations/0011_contactus_feedback.py
sb3958: @AyalaCnaan When a user sends their feedback, the maximum length the database will hold is 2000.

jcarrete5: For technical reasons, [the `CharField` requires a max length be specified](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.CharField.max_length). A [`TextField`](https://docs.djangoproject.com/en/3.2/ref/models/fields/#textfield) does not have this restriction and would probably be better suited for something like the feedback message.

As for why 2000 was specifically chosen... :man_shrugging:

api/models.py
jcarrete5: Looks like the `0011_contactus_feedback.py` creates this model.

api/models.py
```diff
@@ -140,3 +140,17 @@ class ReportType(models.TextChoices):
...
+    name = models.CharField(max_length=50, blank=True)
+    message = models.CharField(max_length=2000, null=True)
+
+
+class Feedback(models.Model):
+    FEEDBACK_REASON = (("W", "Website"), ("Pp", "Penpal"), ("P", "Performance"))
+
+    reason = models.CharField(max_length=2, choices=FEEDBACK_REASON, null=True)
+    message = models.CharField(max_length=2000, blank=True)
```
jcarrete5: I think this would be better suited as a [TextField](https://docs.djangoproject.com/en/3.2/ref/models/fields/#textfield) rather than a CharField since the message could be arbitrarily large.

api/views.py
jcarrete5:
```suggestion
class ContactUs(generics.CreateAPIView):
    serializer_class = ContactUsSerializer
    permission_classes = (permissions.AllowAny,)
```
There are some mistakes in that suggestion that wouldn't have the correct effect if committed directly

frontend/src/FullPageRoutes/AccountsPage.tsx
jcarrete5: This is an error. I don't think it was caught because the event is declared with `any` type.

api/views.py
```diff
@@ -416,3 +418,20 @@ class Accounts(generics.RetrieveAPIView):
...
+        serializer = self.serializer_class(data=request.data)
+        return response.Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+class Feedback(generics.CreateAPIView):
+    serializer_class = FeedbackSerializer
+
+    def post(self, request):
+        serializer = self.serializer_class(data=request.data)
+        return response.Response(serializer.data, status=status.HTTP_201_CREATED)
```
```suggestion
```
This method definition is redundant. This basic implementation is provided by default in `generics.CreateAPIView`

jcarrete5 (changes requested):

api/views.py
```diff
@@ -416,3 +418,20 @@ class Accounts(generics.RetrieveAPIView):
...
         return response.Response(serializer.data, status=status.HTTP_200_OK)
+
+
+class ContactUs(generics.CreateAPIView):
+    serializer_class = ContactUsSerializer
+    permission_clsses = permissions.AllowAny
+
+    def post(self, request):
+        serializer = self.serializer_class(data=request.data)
+        return response.Response(serializer.data, status=status.HTTP_201_CREATED)
```
```suggestion
```
This method is redundant for the same reason it is redundant in the feedback view

api/views.py
sb3958: What will happen using this suggestion?

jcarrete5: The mistake appears to have already been committed. `permission_clsses` is misspelt. It should be `permission_classes` like in my revised suggestion. And the value should be a tuple. The comma is important after the `permissions.AllowAny` in the parentheses.

api/views.py
sb3958: Since it is redundant, if I remove this what should I do then?

jcarrete5: You don't need to do anything except remove it. Removing the method should have the same effect as if it were there. Take a look at the `generics.CreateAPIView` source and you will find that the default implementation defines a post method that simply passes the data from the request to the serializer specified by the `serializer_class` class attribute and then saves the data.

api/models.py
```diff
@@ -145,12 +145,12 @@ class ReportType(models.TextChoices):
...
-    message = models.CharField(max_length=2000, null=True)
+    message = models.TextField()
 
 
 class Feedback(models.Model):
     FEEDBACK_REASON = (("W", "Website"), ("Pp", "Penpal"), ("P", "Performance"))
 
-    reason = models.CharField(max_length=2, choices=FEEDBACK_REASON, null=True)
-    message = models.CharField(max_length=2000, blank=True)
+    reason = models.TextField(choices=FEEDBACK_REASON)
```
jcarrete5: This was fine as a `CharField` because it is only meant to store the two-character feedback reason. You might run into issues when you generate migrations if you change from `null=True` to `null=False` (defaults to `null=False`).
```suggestion
    reason = models.CharField(max_length=2, choices=FEEDBACK_REASON)
```

jcarrete5 (changes requested):
Because changes have been made to some models, you will need to generate migrations again.

.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-    "python.pythonPath": "c:\\Users\\Sharelle\\Desktop\\Senior Design\\conteo\\venv\\Scripts\\python.exe"
+    
+    "workbench.editorAssociations": [
+    
+    ]
```
jcarrete5: You need to run `git rm -r --cached .vscode/` and commit to fully remove the `.vscode/` directory from the repository.

jcarrete5 (changes requested):

elp49 (approved):

jcarrete5 (approved):

mneedleman (approved):

PR #87, opened 2021-04-19
Video Editing

mneedleman:
This is a mess, largely in part to the design of the `vidar` library. See more [here](https://github.com/clabe45/vidar).
Known bugs:
- Regardless of whether the video finishes playing, the "stop" button must be pressed in order to view or send.
- Successive trims edit the original video, not the trimmed video. Not sure if this is an issue with vd or us

image


rao53:
Would it be possible to get a walkthrough on this during our meeting tomorrow?

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -6,43 +6,53 @@ import { parseIdentity, request } from "../utils/fetch"
...
 
 const useStyles = makeStyles({
   video_root: {
     backgroundColor: "black",
   },
 })
 
-function sendVideo(
-  blob: Nullable<Blob>,
+export function sendVideo(
```
elp49: where else will `sendVideo` be used that it needs to be exported?

frontend/src/components/video/editing/EditableVideo.tsx
```diff
@@ -0,0 +1,165 @@
...
+  setMovie,
+}: EditableVideoProps): JSX.Element {
+  const canvasRef = React.useRef(document.createElement("canvas"))
+  const videoRef = React.useRef(document.createElement("video"))
+  const progressRef = React.useRef(document.createElement("input"))
+  const [timeRange, setTimeRange] = React.useState<number[]>([0, 0])
+  const [activeTool, setActiveTool] = React.useState<Nullable<ActiveToolType>>(
+    null
+  )
+  const editHistoryRef = React.useRef<Nullable<ActiveToolType>[]>([])
```
elp49: will null ever be pushed to this?

elp49 (commented):
I trimmed a video, saved it, and then trimmed again. The second trim resets my original trim so I am trimming the original video again rather than trimming my already trimmed video. Is that the expected behavior or something to fix down the line?

frontend/src/FullPageRoutes/RecordPage.tsx
mneedleman: You're right, it doesn't need to be

frontend/src/components/video/editing/EditableVideo.tsx
mneedleman: Doesn't seem like it

mneedleman:
> 
> 
> I trimmed a video, saved it, and then trimmed again. The second trim resets my original trim so I am trimming the original video again rather than trimming my already trimmed video. Is that the expected behavior or something to fix down the line?

It's not expected behavior. I'm not sure where the bug is, but I'll look into it. For now, I'll add it to the bugs

AyalaCnaan (dismissed):
In thinking about this, the trimming issues are a low priority. The functionality itself is priority 2, and the ability to trim doesn't impact anything else (except undo edit, which is also not that important). I don't really care if this remains buggy through the end of the term, as long as it doesn't affect other functionalities. 
From my view we have more important things to make sure are working. So frankly, make this a low priority for fixing. We have proof of concept.

frontend/src/FullPageRoutes/RecordPage.tsx
```diff
@@ -17,32 +18,41 @@ const useStyles = makeStyles({
...
+            parser: parseIdentity,
+          }).then(() => history.goBack())
+        }
+      })
   }
 }
 
 export default function RecordPage(): JSX.Element {
-  const [videoBlob, setVideoBlob] = React.useState<Nullable<Blob>>(null)
+  const [, setVideoBlob] = React.useState<Nullable<Blob>>(null)
```
jcarrete5: Why isn't the state value destructured here? What use is a state if the state data isn't referenced anywhere?

frontend/src/utils/movie.tsx
```diff
@@ -0,0 +1,31 @@
...
+  type?: string
+  video?: boolean
+  audio?: boolean
+  mediaRecorderOptions?: Record<string, unknown>
+}
+export function recordMovie(movie: vd.Movie, options: RecordOptions): void {
+  movie.record(options).then((blob: Blob) => {
+    console.log(blob)
+  })
+}
```
jcarrete5: Can you clarify why this is here? It appears to be used nowhere.

rao53: ^^

rao53 (dismissed):
Apart from the bugs you mentioned earlier, I don't see anything else not working as expected..

jcarrete5 (commented):
It looks like the merge commit didn't actually merge anything from master. Updating the branch locally resulted in a fast-forward and nothing changed

frontend/src/FullPageRoutes/RecordPage.tsx
mneedleman: Taking a second look at this, it looks like that state is redundant

frontend/src/utils/movie.tsx
mneedleman: It was written and promptly forgotten about

frontend/src/components/video/editing/EditControls.tsx
```diff
@@ -0,0 +1,41 @@
...
+          />
+        )
+      }, controlRecord)}
+      <EditControl
+        key="editcontrol-undo"
+        selected={false}
+        type="undo"
+        setActiveTool={setActiveTool}
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        onClick={() => {}}
```
elp49, line 37: I forget what you said about the undo feature, will it not be included in this PR?

jcarrete5 (approved):

frontend/src/components/video/editing/EditControls.tsx
mneedleman: If you look in `EditControl`, the `onClick` method sets the tool type and whatever is additionally passed in. Since undo has no additional UI, it doesn't need any extra action. I could have made the prop optional, but this seemed simpler for 1 case

mneedleman: So yes, undo is in here

rao53 (approved):

PR #89, opened 2021-04-23
Missing Modals

rao53:


rao53:
Screen Shot 2021-04-23 at 1 22 48 PM
Screen Shot 2021-04-23 at 11 34 55 AM

rao53:
To be honest, my audit of many of the currently implemented features don't require confirmation modals. The two images above are all I figured were needed. @Conteo-Inc/dev-team  If there are any more, please let me know. It's fairly simple to implement and should only take me a few mins.

elp49:
> To be honest, my audit of many of the currently implemented features don't require confirmation modals. The two images above are all I figured were needed. @Conteo-Inc/dev-team If there are any more, please let me know. It's fairly simple to implement and should only take me a few mins.

a failed signup and the profile cancel edit button are missing modals as well

frontend/src/components/LoginForm.tsx
```diff
@@ -60,9 +62,33 @@ export default function LoginForm({
   const history = useHistory()
 
   const { from } = location.state || { from: { pathname: "/" } }
+  const [isOpen, setOpen] = React.useState<boolean>(false)
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [type, setType] = useState<NotificationType["type"]>("error")
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [message, setMessage] = useState(
+    "Email and/or password entered incorrectly"
+  )
```
elp49: why not set the type and message when you catch the error?

elp49 (dismissed):

frontend/src/components/LoginForm.tsx
rao53: I could.. Lowkey wouldn't make much difference but I'll have that done.

rao53:
> > To be honest, my audit of many of the currently implemented features don't require confirmation modals. The two images above are all I figured were needed. @Conteo-Inc/dev-team If there are any more, please let me know. It's fairly simple to implement and should only take me a few mins.
> 
> a failed signup and the profile cancel edit button are missing modals as well

I'm not sure what a failed signup would necessarily constitute except the sign-up request fails which I can add. I'll look into the profile cancel edit button rn.

frontend/src/components/LoginForm.tsx
mneedleman: Instead of the eslint disable comments, just replace `const [foo, setFoo] = ...` with `const [foo] = ...`

mneedleman (changes requested):
One tiny suggestion

rao53:
Screen Shot 2021-04-25 at 12 30 13 PM
Screen Shot 2021-04-25 at 12 05 58 PM

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -352,6 +363,13 @@ export default function EditableProfileContent({
                   Cancel
                 </Button>
               </Grid>
+              <AbstractModal
+                isModalOpen={isModalOpen}
+                handleConfirm={handleConfirm}
+                handleCancel={handleClose}
+                title="Confirmation"
+                description={`You cant undo this action`}
```
mneedleman: can't

mneedleman (changes requested):

mneedleman (approved):

frontend/src/components/profile/EditableProfileContent.tsx
```diff
@@ -352,6 +363,13 @@ export default function EditableProfileContent({
                   Cancel
                 </Button>
               </Grid>
+              <AbstractModal
+                isModalOpen={isModalOpen}
+                handleConfirm={handleConfirm}
+                handleCancel={handleClose}
+                title="Confirmation"
+                description={`You can't undo this action`}
```
elp49, line 371: should this say "Would you like to undo your edits?" instead?

elp49 (changes requested):

frontend/src/components/profile/EditableProfileContent.tsx
rao53: So, what happens here is for when after editing, they press the cancel button which clears anything they just added. The pop up happens only for the cancel button. Pressing the save button in my opinion doesn't need a modal...

elp49: oh ok, I was reading this wrong. Initially I thought it was telling the user that they cannot cancel their edits.

elp49 (approved):

PR #90, opened 2021-04-24
Remove penpals

jcarrete5:
image

### DIFF
--- a/frontend/src/components/dashboard/MailItem.tsx
+++ b/frontend/src/components/dashboard/MailItem.tsx
@@ -10,6 +10,7 @@ import * as React from "react"
 import { Link } from "react-router-dom"
 import { Nullable } from "../../utils/context"
 import ViewVideo from "../video/ViewVideo"
+import AbstractModal from "../AbstractModal"
 
 const useStyles = makeStyles({
   mailItem: {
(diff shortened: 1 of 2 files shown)


frontend/src/components/dashboard/MailItem.tsx
```diff
@@ -52,6 +55,9 @@ export default function MailItem({
       <Typography variant="h6">
         {created_at ? video_date.toLocaleDateString() : "No video"}
       </Typography>
+      <IconButton onClick={() => removePenpal(id)}>
+        <Delete fontSize="large" style={{ color: "#4b5282" }} />
+      </IconButton>
```
mneedleman: We definitely want to confirm before a penpal is removed

frontend/src/components/dashboard/MailItem.tsx
```diff
@@ -24,6 +25,7 @@ export type MailListItem = {
   viewed_at: Nullable<string>
   created_at: string
   id: number
+  removePenpal: (id: number) => void
```
mneedleman, line 29: If we already pass the id in, then `removePenpal` does not need to be a prop on the component

mneedleman (changes requested):

frontend/src/components/dashboard/MailItem.tsx
jcarrete5: The mail list doesn't refresh after the request is made to delete a penpal. I need to either pass setMail down so that I can update the list when the button on the mail item is clicked, or pass a function that calls set mail down.

jcarrete5: I could've given the effect that requests the data a dependency so it would re-request the list but there was no convenient value to use as a dependency. And I'd still need to pass something down to trigger a re-render for that as well.

elp49 (approved):

frontend/src/components/dashboard/MailItem.tsx
mneedleman: Fair enough

mneedleman (approved):

PR #92, opened 2021-05-04
Add report dialog when reporting a match

jcarrete5:
image

### DIFF
--- a/frontend/src/FullPageRoutes/Matching.tsx
+++ b/frontend/src/FullPageRoutes/Matching.tsx
@@ -84,8 +84,15 @@ function MatchControls({
         title="Report"
         description="Optional report message"
         isModalOpen={showReport}
-        handleConfirm={() => onReport(matchId, description)}
-        handleCancel={() => setShowReport(false)}
+        handleConfirm={() => {
+          setDescription("")
+          setShowReport(false)
+          onReport(matchId, description)
+        }}
+        handleCancel={() => {
+          setDescription("")
+          setShowReport(false)
+        }}
       >
         <TextField
           multiline
--- a/frontend/src/FullPageRoutes/Matching.tsx
+++ b/frontend/src/FullPageRoutes/Matching.tsx
@@ -15,6 +15,7 @@ import { Autocomplete } from "@material-ui/lab"
 import { Check, Block, Flag, PlayCircleFilled } from "@material-ui/icons"
 import { request, queryParams } from "../utils/fetch"
 import ViewVideo from "../components/video/ViewVideo"
+import AbstractModal from "../components/AbstractModal"
 
 type EnqueueFunc<T> = (...newItems: T[]) => void
 
@@ -22,7 +23,7 @@ type MatchControlsType = {
   matchId: number
   onAccept: (matchId: number) => void
   onReject: (matchId: number) => void
-  onReport: (matchId: number) => void
+  onReport: (matchId: number, description?: string) => void
 }
 
 type Match = {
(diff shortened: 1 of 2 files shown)


mneedleman (dismissed):

frontend/src/FullPageRoutes/Matching.tsx
```diff
@@ -75,8 +76,23 @@ function MatchControls({
...
+  const [showReport, setShowReport] = React.useState<boolean>(false)
+  const [description, setDescription] = React.useState<string>("")
   return (
     <Grid item container direction="row" justify="space-around">
+      <AbstractModal
+        title="Report"
+        description="Optional report message"
+        isModalOpen={showReport}
+        handleConfirm={() => onReport(matchId, description)}
+        handleCancel={() => setShowReport(false)}
```
elp49: should the description be set to an empty string on cancel or at least after a match has been accepted/rejected?

frontend/src/FullPageRoutes/Matching.tsx
```diff
@@ -75,8 +76,23 @@ function MatchControls({
...
 }: MatchControlsType): JSX.Element {
+  const [showReport, setShowReport] = React.useState<boolean>(false)
+  const [description, setDescription] = React.useState<string>("")
   return (
     <Grid item container direction="row" justify="space-around">
+      <AbstractModal
+        title="Report"
+        description="Optional report message"
+        isModalOpen={showReport}
+        handleConfirm={() => onReport(matchId, description)}
```
elp49: should the report dialog be closed after confirmed?

elp49 (changes requested):

frontend/src/FullPageRoutes/Matching.tsx
jcarrete5: The controls will re-render after the report is submitted because a new match will be requested. No need to explicitly set the modal state to false.

elp49: that didn't happen when I tested it

jcarrete5: Hmm actually looks like it keeps the same state unless there are no more matches

frontend/src/FullPageRoutes/Matching.tsx
```diff
@@ -84,8 +84,15 @@ function MatchControls({
         title="Report"
         description="Optional report message"
         isModalOpen={showReport}
-        handleConfirm={() => onReport(matchId, description)}
-        handleCancel={() => setShowReport(false)}
+        handleConfirm={() => {
+          setDescription("")
+          setShowReport(false)
+          onReport(matchId, description)
```
elp49, line 90: this should probably come before `setDescription("")`

elp49 (changes requested):

frontend/src/FullPageRoutes/Matching.tsx
jcarrete5: why? Pretty sure the order doesn't matter. afaik the state won't change until after the function is finished executing

jcarrete5: And the data in the request appears to confirm my assumption. description is correctly set when making the request.

elp49: oh ok, I didn't realize that

elp49 (approved):

frontend/src/FullPageRoutes/Matching.tsx
jcarrete5: I think it would be more accurate to say the state variables update when the component re-renders.

mneedleman (approved):

PR #93, opened 2021-05-05
Allow filters to be ignored

jcarrete5:
image

Filter can be ignored so users who haven't specified an age, gender, or interest can be matched with people who don't specify those filters


frontend/src/FullPageRoutes/Matching.tsx
```diff
@@ -222,7 +279,9 @@ export default function MatchingPage(): JSX.Element {
         "/api/matches/" +
         queryParams({
           ...filters,
-          interests: filters.interests.map((v) => v.id),
+          interests: filters.interests
+            ? filters.interests.map((v) => v.id)
+            : undefined,
```
mneedleman: I prefer to avoid undefined whenever possible. What do you think of this? I haven't tested it, but it's the same syntax I used in `fetch.tsx`

```suggestion
          ...(filters.interests && {"interests": filters.interests.map((v) => v.id)})
```

frontend/src/utils/fetch.tsx
```diff
@@ -92,6 +92,9 @@ export function queryParams(obj: Record<PropertyKey, unknown>): string {
           throw new Error(`Failed to encode key/value pair: ${key}=${val}`)
         }
         break
+      case "undefined":
+        /* Don't add any query param if value is undefined */
+        break
```
mneedleman, line 97: I believe my suggestion above removes the need for this

mneedleman (changes requested):
Something to consider, otherwise looks good

frontend/src/FullPageRoutes/Matching.tsx
jcarrete5: Yea I think I like your change better.

frontend/src/utils/fetch.tsx
jcarrete5: This is still needed because the min and max age use undefined to signal that a min and max age isn't specified. And in general, I decided that any variable that has a value of undefined should just be ignored rather than error. I suppose I could look at changing `...filters` to something that avoids including properties that are undefined but I don't think it's worth the effort.

jcarrete5: Maybe that Ramda library has something for this? I'm not familiar with it.

mneedleman: Maybe, but for now this is fine

elp49 (approved):
works well

mneedleman (approved):

PR #94, opened 2021-05-05
View Penpal Profile

elp49:
- added button to penpal list item to view profile
- replacing list item account circle icon with penpal profile picture
- added penpal profile retrieve view
- added filter function to remove hidden penpal profile content

Note: in the screenshots below, profile content with privacy setting Hidden are blank


mneedleman (dismissed):

jcarrete5 (changes requested):
Can you merge master into this branch? It looks like there are missing changes in this PR similar to what happened with video editing.

frontend/src/components/MailItem.tsx
```diff
@@ -68,10 +115,30 @@ export default function MailItem({
...
+      </AbstractModal>
       <Grid item container direction="row" xs={3} wrap="nowrap">
-        <AccountCircle fontSize="large" style={{ color: "#4b5e82" }} />
+        <IconButton onClick={() => setShowProfile(true)}>
+          <Avatar
+            src={profileContent.image ? profileContent.image : ""}
+            alt={profileContent.first_name}
+          />
+        </IconButton>
         <Typography variant="h6">{`${first_name} ${last_name}`}</Typography>
```
rao53, line 170: I set the first and last name of one of the profiles as null and it still displayed the name on the dashboard but set it as blank on the view_profile abstract modal. Something to think about: do we even want people setting privacy on their first and last name?

rao53 (changes requested):
Picture below to support my comment. This is test (First name) one (Last name) user
Screen Shot 2021-05-06 at 10 20 17 PM

frontend/src/components/MailItem.tsx
elp49: that's a good point, it would be hard to identify your penpals if multiple have not image and their names are hidden

elp49:
> Can you merge master into this branch? It looks like there are missing changes in this PR similar to what happened with video editing.

merged

frontend/src/utils/profile.tsx
```diff
@@ -216,3 +216,18 @@ export function toDateString(date: Date): string {
...
+    birthday = new Date(birth_date)
+    // We have Django configured to store dates using local time.
+    // Date objects assume the date is UTC when parsing so we add 1
+    // to correct for the EST/EDT - UTC timezone difference
+    birthday.setDate(birthday.getDate() + 1)
+  }
+  console.log("birthday:", birthday)
+
+  return birthday
+}
```
jcarrete5: So I'm not sure what happened but this function should have been deleted. The dates on the backend aren't really aware of timezone since we only store the date and not the time. The `toDateString()` function above interprets the date as UTC and avoids having to add one to adjust for the timezone and should be used in place of `Date.toLocaleDateString()` for displaying the birth date. This function should be removed.

jcarrete5 (changes requested):

frontend/src/components/MailItem.tsx
mneedleman: At a minimum, last name should be able to be hidden

rao53: I agree with at least last name being hidden and first name being public at all times.

frontend/src/utils/profile.tsx
elp49: sounds good

frontend/src/components/MailItem.tsx
```diff
@@ -1,15 +1,25 @@
...
+import {
   Delete,
   DraftsRounded,
   MailOutlineRounded,
   SendRounded,
 } from "@material-ui/icons"
 import ViewVideo from "./video/ViewVideo"
 import { Nullable } from "../utils/context"
+import { request } from "../utils/fetch"
+import { parseBirthday } from "../utils/profile"
```
jcarrete5: Delete the function from `utils/profile.tsx`. This is an unused import.

jcarrete5 (changes requested):

jcarrete5 (dismissed):

frontend/src/components/MailItem.tsx
```diff
@@ -50,6 +59,45 @@ export default function MailItem({
...
+
+  React.useEffect(() => {
+    request<ProfileContentType>({ path: `/api/profiles/${id}/`, method: "get" })
+      .then((res) => {
+        const profile_content = res.parsedBody
+
+        // Fields that are restricted on a user's profile (i.e. Private or Hidden)
+        // are not returned for privacy purposes. Test if certain field were not
+        // returned to avoid frontend errors related to undefined values.
+        if (typeof profile_content.birth_date === "undefined") {
```
rao53, line 86: One thing I found was the birthday of new users was set automatically which is weird. Also, the bug from my last comment is still there. I set the first and last name of a match to hidden and it didn't hide those demographics which were right next to the "view profile" button.

elp49: I'll add a parseBirthday function back in to set birth_date to null if one is not set.
As for your last comment, I created a clickup task to remove privacy setting from first_name.

I can add a check for last_name privacy settings for mail list items unless you, @Deravadom, or @jcarrete5 think that should be in a separate task

elp49: that would hide the last name of a penpal from the mail page if set to hidden

mneedleman (dismissed):

mneedleman (dismissed):

rao53 (dismissed):

mneedleman (dismissed):

PR #95, opened 2021-05-05
Viewed Penpal Video

elp49:
recording when a video has been viewed


api/views.py
```diff
@@ -289,6 +289,17 @@ def get(self, request):
...
+    def put(self, request, video_id):
+        video_object = Video.objects.get(id=video_id)
+        viewed_at = datetime.now()
+        video_object.viewed_at = viewed_at
+        video_object.save()
+        return response.Response(
+            data={"viewed_at": str(viewed_at)}, status=status.HTTP_200_OK
+        )
+
+
```
mneedleman: Why not combine with `MailListView` to make `MailUpdateListView`

mneedleman (changes requested):

api/views.py
jcarrete5: If the client instead sends the `viewed_at` date, then it would be fairly straight forward to combine the two views.
You'd only need to add the `ListModelMixin` and override the `get_object` method to return the video with the specified ID.

I don't think this is worth the effort though. The endpoints need to be heavily refactored imo.

mneedleman: Fair enough, and I agree

jcarrete5 (dismissed):
Regarding organization, I think it would be better to have a serializer handle updating the Video table in the `MailUpdateView` but the code works fine as-is.

jcarrete5 (dismissed):

mneedleman (approved):

jcarrete5 (approved):

PR #101, opened 2021-05-10
Header/Footer display on top of everything else

elp49:

### DIFF
--- a/frontend/src/components/core/AppFooter.tsx
+++ b/frontend/src/components/core/AppFooter.tsx
@@ -10,6 +10,7 @@ const useStyles = makeStyles({
     position: "fixed",
     left: "0px",
     bottom: "0px",
+    zIndex: 1,
   },
   bannerText: {
     color: "white",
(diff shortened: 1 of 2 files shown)


mneedleman (approved):
This error is symptomatic of a larger issue, and adding a z-index is just a band-aid. But given the time constraint, this will have to do.

rao53 (approved):

rao53:
> This error is symptomatic of a larger issue, and adding a z-index is just a band-aid. But given the time constraint, this will have to do.

What would be a good alternative?

mneedleman:
> 
> 
> > This error is symptomatic of a larger issue, and adding a z-index is just a band-aid. But given the time constraint, this will have to do.
> 
> What would be a good alternative?

The issue lies with overlapping grids. Making them display on top just moves them out of the way, but the body of the app will still have issues

elp49:
> > > This error is symptomatic of a larger issue, and adding a z-index is just a band-aid. But given the time constraint, this will have to do.
> > 
> > 
> > What would be a good alternative?
> 
> The issue lies with overlapping grids. Making them display on top just moves them out of the way, but the body of the app will still have issues

from what I've seen, this happens when `position:relative` and `position:fixed/absolute` clash

PR #102, opened 2021-05-12
Remove Profile Sidebar

elp49:
- removed profile sidebar
- deleted profile privacy settings component
- removed privacy stuff from profile page
- made privacy settings its own page and changed profile/privacy views accordingly


mneedleman:
Why does a PR about removing the extra sidebar have a bunch of extra changes? That's not how we do things

elp49:
> Why does a PR about removing the extra sidebar have a bunch of extra changes? That's not how we do things

what changes are you talking about?

mneedleman (dismissed):
I'm concerned about the scope of this, but it'll have to do

mneedleman (approved):

PR #105, opened 2021-05-14
easy fix

rao53:
API request was retrieving all videos in the db instead.


api/tests/test_video.py
```diff
@@ -51,8 +51,8 @@ def test_send_video(self):
         self.client.login(username="lo", password="lo")
         res = self.client.get("/api/videos/")
         results = res.json()
-        self.assertEqual(len(results), 1)
-        self.assertEqual(results[0]["video_file"], self.video)
+        self.assertEqual(len(results), 2)
+        self.assertEqual(results[1]["video_file"], self.video)
```
jcarrete5: Is this the desired behavior though? If I log in as a user and make a request to `/api/videos/`, is the user supposed to get every video in the database or just the videos sent to them?

mneedleman:
Wait if this is a bug in the views, why are we fixing it in the test? I'm confused about what's going on here

mneedleman (approved):

